Issue #1, opened 2017-10-16
initialize ksonnet migration


jessicayuen:
- [x] **PR #2** rename commands to `ks`
- [x] **PR #2** rename packages to `ksonnet`
- [x] migrate ksonnet.next issues
- [x] migrate ksonnet.next PRs
- [x] setup CI

PR #2, opened 2017-10-16
Rename binary to 'ks';  rename 'kubecfg' references to 'ksonnet'

jessicayuen:
As a part of the migration work from [github.com/ksonnet/kubecfg](github.com/ksonnet/kubecfg) to this repository, we will change the naming of this command line tool to 'ksonnet'. This implies the following changes:

- Calling the executable 'ks', so that we can run commands such as `ks apply dev`
- Rename all references of 'kubecfg' in the codebase to 'ksonnet'.

Partially supports issue #1.


jessicayuen:
> So. I'd like to propose that we only change the public-facing stuff instead of the internal APIs, because after kubecon we're going to have to pull the common stuff out and merge it back into a common repository, and we don't want to make life harder than it has to be.

Fixed this -- only binary name change (+docs) and import path fixes.

Issue #7, opened 2017-10-16
Support `diff` between two environments


jessicayuen:
_From @jessicayuen on September 27, 2017 20:11_

From [spec](https://docs.google.com/document/d/1yU6Zi3t1D-sSLnMEE6aQn2NS41tLqiYgzEoZnQvQEbY):

`ksonnet diff <env-one> [env-two]`

Diff should model git’s diff between two commits / branches. In the case of environments, diff will output the changes between the components deployed at the API server located at `<env-one>` and the API server located at `<env-two>`. `<env-#>` is the name of an environment. Environments must be created using `ksonnet env add` before use.

Examples:
  Output the differences in components deployed at remote and local “us-west/dev”.
   `ksonnet diff us-west/dev`

  Output the differences in components deployed at remote “us-west/dev” and remote “us-west/prod” environments.
   `ksonnet diff remote:us-west/dev remote:us-west/prod`

  Output the differences in components deployed at local “us-west/dev” and remote “us-west/prod” environments.
   `ksonnet diff local:us-west/dev remote:us-west/prod`

_Copied from original issue: ksonnet/kubecfg#163_

jessicayuen:
_From @seh on September 30, 2017 13:10_

Does _diff_ elide details like creation time, generation, and other controller-set fields?

jessicayuen:
Not currently, but we've had conversations regarding excluding certain fields. There's some concerns around the maintainability of this list.

PR #19, opened 2017-10-16
Support `diff`ing between two environments

jessicayuen:
**SEE [PR](https://github.com/ksonnet/kubecfg/pull/167) FROM OLD REPOSITORY FOR EXISTING COMMENTS**

**Fixes: #7**

First look at supporting diff between two environments.

This PR will support changes such as
`ks diff local:dev local:prod`,
`ks diff remote:dev remote:prod`, or
`ks diff local:dev remote:dev`. 

The last is the same as the existing diff functionality `ks diff dev` (which is still supported).

### DIFF
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -29,7 +29,6 @@ const flagDiffStrategy = "diff-strategy"
 
 func init() {
 	addEnvCmdFlags(diffCmd)
-	bindClientGoFlags(diffCmd)
 	bindJsonnetFlags(diffCmd)
 	diffCmd.PersistentFlags().String(flagDiffStrategy, "all", "Diff strategy, all or subset.")
 	RootCmd.AddCommand(diffCmd)
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -18,8 +18,14 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
+
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/ksonnet/ksonnet/metadata"
 	"github.com/ksonnet/ksonnet/pkg/kubecfg"
@@ -35,22 +41,14 @@ func init() {
 }
 
 var diffCmd = &cobra.Command{
-	Use:   "diff [env-name] [-f <file-or-dir>]",
-	Short: "Display differences between server and local config",
+	Use:   "diff [<env1> [<env2>]] [-f <file-or-dir>]",
+	Short: "Display differences between server and local config, or server and server config",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) > 1 {
-			return fmt.Errorf("'diff' takes at most a single argument, that is the name of the environment")
+		if len(args) > 2 {
+			return fmt.Errorf("'diff' takes at most two arguments, that are the name of the environments")
 		}
 
 		flags := cmd.Flags()
-		var err error
-
-		c := kubecfg.DiffCmd{}
-
-		c.DiffStrategy, err = flags.GetString(flagDiffStrategy)
-		if err != nil {
-			return err
-		}
 
 		cwd, err := os.Getwd()
 		if err != nil {
@@ -58,36 +56,55 @@ var diffCmd = &cobra.Command{
 		}
 		wd := metadata.AbsPath(cwd)
 
-		envSpec, err := parseEnvCmd(cmd, args)
+		files, err := flags.GetStringArray(flagFile)
 		if err != nil {
 			return err
 		}
 
-		c.ClientPool, c.Discovery, err = restClientPool(cmd, envSpec.env)
+		var env1 *string
+		if len(args) > 0 {
+			env1 = &args[0]
+		}
+
+		var env2 *string
+		if len(args) > 1 {
+			env2 = &args[1]
+		}
+
+		diffStrategy, err := flags.GetString(flagDiffStrategy)
 		if err != nil {
 			return err
 		}
 
-		c.Namespace, err = namespace()
+		c, err := initDiffCmd(cmd, wd, env1, env2, files, diffStrategy)
 		if err != nil {
 			return err
 		}
 
-		objs, err := expandEnvCmdObjs(cmd, envSpec, wd)
-		if err != nil {
-			return err
-		}
-
-		return c.Run(objs, cmd.OutOrStdout())
+		return c.Run(cmd.OutOrStdout())
 	},
-	Long: `Display differences between server and local configuration.
+	Long: `Display differences between server and local configuration, or server and server
+configurations.
 
 ksonnet applications are accepted, as well as normal JSON, YAML, and Jsonnet
 files.`,
-	Example: `  # Show diff between resources described in a local ksonnet application and
-  # the cluster referenced by the 'dev' environment. Can be used in any
-  # subdirectory of the application.
-  ks diff dev
+	Example: `  # Show diff between resources described in a the local 'dev' environment
+  # specified by the ksonnet application and the remote cluster referenced by
+  # the same 'dev' environment. Can be used in any subdirectory of the application.
+  ksonnet diff dev
+
+  # Show diff between resources at remote clusters. This requires ksonnet
+  # application defined environments. Diff between the cluster defined at the
+  # 'us-west/dev' environment, and the cluster defined at the 'us-west/prod'
+  # environment. Can be used in any subdirectory of the application.
+  ksonnet diff remote:us-west/dev remote:us-west/prod
+
+  # Show diff between resources at a remote and a local cluster. This requires
+  # ksonnet application defined environments. Diff between the cluster defined
+  # at the 'us-west/dev' environment, and the cluster defined at the
+  # 'us-west/prod' environment. Can be used in any subdirectory of the
+  # application.
+  ksonnet diff local:us-west/dev remote:us-west/prod
 
   # Show diff between resources described in a YAML file and the cluster
   # referenced in '$KUBECONFIG'.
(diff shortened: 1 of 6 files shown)


pkg/kubecfg/diff.go
```diff
@@ -193,19 +193,24 @@ func diff(desc, aName, bName, strategy string, aObj, bObj map[string]interface{}
 	return false, nil
 }
 
-func hash(discovery *discovery.DiscoveryInterface, obj *unstructured.Unstructured) string {
+// hash serves as an identifier for the Kubernetes resource.
+func hash(discovery *discovery.DiscoveryInterface, obj *unstructured.Unstructured, fqName bool) string {
```
hausdorff: Does the `fqName` stuff belong in the commit `Remove client-go bindings from cmd`? It seems like it should just be removing the `client-go` flags.

Also, I think these commits are out of order?

cmd/diff.go
```diff
@@ -36,22 +38,14 @@ func init() {
 }
 
 var diffCmd = &cobra.Command{
-	Use:   "diff [env-name] [-f <file-or-dir>]",
-	Short: "Display differences between server and local config",
+	Use:   "diff [env-name] [env-name] [-f <file-or-dir>]",
```
hausdorff: Use here is probably: `diff <env1> [env2] [-f <file-or-dir>]`. Not sure how informative that is without the examples though.

cmd/diff.go
```diff
@@ -102,3 +105,99 @@ files.`,
...
+
+	// ---------------------------------------------------------------------------
+	// Diff between expanded Kubernete objects and objects on a remote cluster
+	// ---------------------------------------------------------------------------
+	if envSpec.env2 == nil {
+		c := kubecfg.DiffRemoteCmd{}
+		c.DiffStrategy = diffStrategy
+		c.Client = &kubecfg.Client{}
+
+		c.Client.APIObjects, err = expandEnvCmdObjs(cmd, envSpec, wd)
```
hausdorff: I'm not quite sure this is sufficient. If you do something like `ks diff local:default` we'll actually look for the environment called `local:default`, but I think we actually want to error out and report that `local:` isn't allowed in a `diff` with a single environment.

cmd/diff.go
```diff
@@ -102,3 +105,99 @@ files.`,
...
+
+	env1 := strings.SplitN(*envSpec.env, ":", 2)
+	env2 := strings.SplitN(*envSpec.env2, ":", 2)
+
+	if len(env1) < 2 || len(env2) < 2 || (env1[0] != local && env1[0] != remote) || (env2[0] != local && env2[0] != remote) {
+		return nil, fmt.Errorf("[env-name] must be prefaced by %s: or %s:, ex: %s:us-west/prod", local, remote, remote)
+	}
+
+	// ---------------------------------------------------------------------------
+	// Diff between two sets of expanded Kubernete objects
```
hausdorff: `Kubernete` is a misspelling. :)

cmd/root.go
```diff
@@ -261,7 +261,9 @@ func restClientPool(cmd *cobra.Command, envName *string) (dynamic.ClientPool, di
 }
 
 type envSpec struct {
-	env   *string
+	env  *string
+	env2 *string
```
hausdorff: I'm not sure this is the best place for `env2`. `envSpec` was semantically supposed to specify the files of a single environment. With two environment names, I'm not sure what `envSpec` "means" anymore.

cmd/diff.go
```diff
@@ -102,3 +105,99 @@ files.`,
...
+
+		c.Client.Namespace, err = namespace()
+		if err != nil {
+			return nil, err
+		}
+
+		return &c, nil
+	}
+
+	env1 := strings.SplitN(*envSpec.env, ":", 2)
```
hausdorff: I think we might need to disallow the `:` character in environment identifiers, because of the example above, `ks diff local:default`, it gets confusing for users to reason about it. Here are the options I see:

1. Environment IDs can't begin with the prefix `local:`, which makes `ks diff local: default` invalid, but then we're in the weird position of disallowing a single, specific prefix in environment IDs, or
1. Environment IDs can have `:` in them and the prefix `local:` _is_ allowed, in which which `ks diff local:default` parses to look for the environment _called_ `local:default`, which is different from the _local version of the environment called `local:default`_ (which would be denoted as `local:local:default`).
1. Environment IDs can't have `:` in them, which is more restrictive, but less confusing in every case I can think of.

I don't know what you think, but I like the last option most.

pkg/kubecfg/diff.go
```diff
@@ -35,24 +35,60 @@ import (
 
 var ErrDiffFound = fmt.Errorf("Differences found.")
 
-// DiffCmd represents the diff subcommand
-type DiffCmd struct {
+// DiffCmd is an interface containing a set of functions that allow diffing
+// between two sets of data containing Kubernete resouces.
```
hausdorff: Another misspelling `Kubernete`.

cmd/diff.go
jessicayuen: Going to follow the `git diff` schematics and do `diff [<env1> [<env2>]] [-f <file-or-dir>] to also capture this case: `ks diff -f ./pod.yaml`

pkg/kubecfg/diff.go
jessicayuen: urgh, I think I did a bad rebase. Thanks for catching.

cmd/diff.go
jessicayuen: We already don't allow for punctuations in env names, which includes `:`.  I added a check to `ks diff local:env`

cmd/root.go
jessicayuen: Hmm, what do you propose here? This is the sort of place where I wish go had options, but instead we have pointers.

On another note though, I'm not sure if that's what `envSpec` semantically specified even prior to this change.  The `files` in envSpec doesn't correspond to the env files (i.e. the expanded `dev.jsonnet`), but instead the contents of `-f` flag (empty if not provided).

hausdorff: Ok, so when I say that "`envSpec` was semantically supposed to specify the files of a single environment", what I mean is that it's supposed to semantically represent the files for commands that take either _a set of `-f` flags_, or an _env name_, or _both_.

Adding a second environment makes every command that uses this data structure more complicated, because it has to know that `env2` is not important.

So fundamentally I think the problem here is that `parseEnvCmd` does not do what you want, because `diff` is no longer an "env command". What `diff` really wants is: (1) a single environment, (2) two environments, (3) a single environment with a `-f`, or (4) two environments with a `-f` flag. So in `diff` I'd just parse that logic.

It isn't awesome, but no one else needs this to be a general part of the abstraction.

jessicayuen: Makes sense, I'll make some changes. Just to clarify -- `diff` is true for the first 3 cases that you specified, but what is the behavior of "(4) two environments with a `-f` flag?" I don't think that case should be supported because it appears to be a 3-way diff.

hausdorff: Is it a three-way merge? A command like `ks diff local:foo local:bar -f components/whatever` should diff those environments, but _only_ for that file, is what I was thinking.

jessicayuen: Follow up question would then be -- is `ks diff foo -f guestbook.jsonnet` diffing the resources in `guestbook.jsonnet` between the equivalent of `local:foo` or `remote:foo`?

cmd/diff.go
```diff
@@ -102,3 +118,183 @@ files.`,
...
+		c.DiffStrategy = diffStrategy
+		c.Client = &kubecfg.Client{}
+
+		if envFq1 != nil && (strings.HasPrefix(*envFq1, remote+":") || strings.HasPrefix(*envFq1, local+":")) {
+			return nil, fmt.Errorf("single <env> argument with prefix '%s': or '%s': not allowed", local, remote)
+		}
+
+		envSpec := &envSpec{env: envFq1, files: files}
+
+		c.Client.APIObjects, err = expandEnvCmdObjs(cmd, envSpec, wd)
```
hausdorff: Ok, so I think the pattern here is that these three types of diff involve some combination of the following two cases:

1. Expand local version of the app into Kubernetes API objects, pass those to `kubecfg.*`
1. Get a client pool, pass to `kubecfg.*` and have that retrieve the API objects from the server.

In other commands, the API objects are always expanded before passing to `kubecfg.*`, and the reason is to make the interfaces testable. (You may remember Gus complained and we had to basically revert.) This will become more important as we have more unit tests to substantiate diffs based on the persistent cache of previous app states.

So: I suspect but do not know that it would be simpler to have a single `kubecfg.DiffCmd` type that contains two sets of API objects and a diff strategy, and a function `Run` that simply executes the diff, which would be much more similar in principle to how (_e.g._) `apply` works.

Assuming this makes sense, I would also break this monofunction into smaller sub-functions (as mentioned above).

jessicayuen: I'll break up `initDiffCmd()` into smaller functions but i'm not sure that having a single `kubecfg.DiffCmd` taking two sets of API objects and a diff strategy is the right option for a few reasons --

1. We'd be coupling the logic for retrieving live objects further with the `cmd/` package, which is already doing much more than adding commands and command line parsing.
2. If it's the unit testability of diffing between two sets of API objects that is a concern, testing the `diffAll` function should capture this. It's a standalone function performing the diff that takes two sets of API objects as it's main params. 
3. I worry we wouldn't be capturing the state of things -- that we are in fact supporting multiple diff command types; diffing between objects locally, between objects on server, and between objects on server and locally. I feel that the application logic layer should be capturing this in the `kubecfg` package instead of hidden in `cmd`.

cmd/diff.go
```diff
@@ -29,66 +35,76 @@ const flagDiffStrategy = "diff-strategy"
...
-		}
 
 		cwd, err := os.Getwd()
 		if err != nil {
 			return err
 		}
 		wd := metadata.AbsPath(cwd)
 
-		envSpec, err := parseEnvCmd(cmd, args)
+		files, err := flags.GetStringArray(flagFile)
```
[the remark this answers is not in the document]
jessicayuen: Added a check to error out on this case.

cmd/diff.go
hausdorff: Ok, it turns out `kubecfg.DiffCmd` is currently fetching from the server, so I'm fine with this staying as it is.

pkg/kubecfg/diff.go
```diff
@@ -35,70 +35,202 @@ import (
...
+// DiffLocalCmd extends DiffCmd and is meant to represent diffing between two
+// sets of Kubernetes objects.
+type DiffLocalCmd struct {
+	Diff
+	Env1 *LocalEnv
+	Env2 *LocalEnv
+}
+
+func (c *DiffLocalCmd) Run(out io.Writer) error {
+	sort.Sort(utils.AlphabeticalOrder(c.Env1.APIObjects))
```
hausdorff: Probably should move this sort to diffAll.

pkg/kubecfg/diff.go
```diff
@@ -35,70 +35,202 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	return diffAll(liveObjsA, liveObjsB, c.ClientA.Name, c.ClientB.Name, c.DiffStrategy, &c.ClientA.Discovery, false, out)
+}
+
+// ---------------------------------------------------------------------------
+
+func diffAll(a []*unstructured.Unstructured, b map[string]*unstructured.Unstructured, aName, bName, strategy string,
```
hausdorff: Ha ha, what a beast of a function signature.

PR #24, opened 2017-10-19
 Init env metadata from provided context

### LINKED ISSUE #22: Init env metadata from provided context
Automatically init namespace and URI from some context. `ks env add us-west/staging --context your-context`

jessicayuen:
Fixes #22 

`ks env add myenv --context=dev` will initialize the environment URI 
and namespace based on the context `dev`, as specfied in the
kubeconfig file. The same applies for `ks init foo --context=dev`.

If both the context and uri argument is not provided, the current
context is used.

This change will also remove uri as a mandatory arg in `ks env add`.
The uri will be moved to a flag. It cannot be used at the same time as
`context`.


cmd/env.go
```diff
@@ -95,18 +100,45 @@ var envAddCmd = &cobra.Command{
 	Short: "Add a new environment to a ksonnet project",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		flags := cmd.Flags()
-		if len(args) != 2 {
-			return fmt.Errorf("'env add' takes two arguments, the name and the uri of the environment, respectively")
+		if len(args) < 1 {
+			return fmt.Errorf("'env add' takes at least one argument, which is the name of the environment")
+		}
+		if len(args) > 2 {
+			return fmt.Errorf("'env add' takes at most two arguments, which is the name and the uri of the environment, respectively")
```
hausdorff: I am thinking that we might want the second argument to be context name instead of URI. Then we'd allow people to tweak the environment with `--uri`, `--namespace`, `--context`, and so on. So the form would be something like:

```
$ ks init <app-name> [ctx-name] [--kubefile=<kf>] [--namespace=<ns>] [--uri=<uri>].
```

You wouldn't need a context if you have a `--uri`.

This complicates the logic somewhat, but I think people generally do not want to pass in the URI directly. I don't think most of our users will even know where it is.

jessicayuen: It's a redesign from the original intention of the issue, but I agree with you -- I think it'll be clearer for users to understand how to initiate environments using the context.

Changing the command line logic is relatively straightforward, but how far do you think we should deviate from the original design -- i.e., should the envrionment's `spec.json` be storing the cluster URI or the context?

jessicayuen: We talked offline. Decided to change the command line interface for `ks env add` and `ks init` to take optional `--context` and `--uri` flags, but the logic around storing env metadata, i.e. storing `uri` in `spec.json`, will remain unchanged.

cmd/env.go
```diff
@@ -293,6 +332,33 @@ the name of an environment will also update the directory structure in
...
+		return "", "", "", err
+	}
+
+	context, err = flags.GetString(flagEnvContext)
+	if err != nil {
+		return "", "", "", err
+	}
+
+	if len(context) != 0 && len(uri) != 0 {
+		return "", "", "", fmt.Errorf("command cannot take both '%s' and '%s' flags", context, uri)
```
hausdorff: Looks like you're using the value of these flags, so an error message would look something like:

```
command cannot take both 'some-ctx' and 'example.com' flags
```

Seems like you probably want something like:

```
command cannot take both '--context' and '--uri' flags
```

Also if we keep this behavior, we probably might want to explain why. Perhaps something like:

```
Flags '--context' and '--uri' are mutually exclusive, because '--context' has a URI. Try setting '--uri', '--namespace' to the desired values.
```

cmd/env.go
```diff
@@ -293,6 +332,33 @@ the name of an environment will also update the directory structure in
...
-  ks env set us-west/staging --uri=http://example.com --name=us-east/staging`,
+  # Updating the name will update the directory structure in 'environments'.
+  ks env set us-west/staging --uri=http://example.com --name=us-east/staging
+  
+  # Updates URI of the environment 'us-west/staging' based on the server URI
+  # in the context 'staging-west' in your kubeconfig file.
+  ks env set us-west/staging --context=staging-west`,
+}
+
+func commonEnvFlags(flags *pflag.FlagSet) (uri, namespace, context string, err error) {
```
hausdorff: Two things.

1. I know I suggested earlier that we should make `--context` and `--uri` mutually exclusive, but now I'm not so sure. Now I'd like to propose that we do whatever kubectl does. What do you think? Sorry for being slightly schizoid about this. I keep forgetting that we have a reference implementation for these flags.
1. If we keep this mutual exclusion, I think it's worth mentioning a comment of this function that `--uri` and `--context` are mutually exclusive, but `--context` and `--namespace` are not, as well as _why_.

cmd/env.go
```diff
@@ -75,38 +83,61 @@ uniquely identifies the cluster.
...
-		envURI := args[1]
+		name := args[0]
 
-		envNamespace, err := flags.GetString(flagEnvNamespace)
+		uri, namespace, context, err := commonEnvFlags(flags)
 		if err != nil {
 			return err
 		}
 
+		if len(uri) == 0 {
```
hausdorff: This is an interesting bit of code because the mutual exclusion of the `--uri` and `--context` flags is baked into the semantics. I'd personally prefer that we hide the call to `server` behind `commonEnvFlags`, and guaranteeing that we always return a valid triple of {uri, ns, ctx}. So, basically, all the logic of this conditional would probably be in `commonEnvFlags`.

Thoughts?

cmd/env.go
jessicayuen: 1. So, client-go doesn't have a flag `uri`, but it has a `server` flag. It doesn't complain if we try to provide both flags but the server flag seems to be ignored.
2. Using your message from below.

cmd/env.go
jessicayuen: I had the call to `server()` hidden in `commonEnvFlags` originally, but had to pull it out because of some fundamental difference in behaviors between `env set` and env add`.

i.e.,
if the `--context` and `--uri` flags are not provided for `env add`, we want to use the default context, pulled from kubeconfig. However, for `env set`, it just means the user doesn't want to update their environment's uri-related metadata. It becomes not obvious when `commonEnvFlags` should call `server()` since it is command-agnostic.

Let me know if you see an alternative.

cmd/env.go
hausdorff: Ok, so we should probably have a `--server` flag, then, is what I'd say.

jessicayuen: Talked offline, we're going to punt this to a follow-up PR to change `uri` references to `server` across the board, including the environment's `spec.json` file.

cmd/env.go
hausdorff: Ok, I think this is not a thing I'm prepared to hold up the review for.

cmd/root.go
```diff
@@ -128,6 +128,28 @@ func namespace() (string, error) {
 	return ns, err
 }
 
+// server returns the server URI and namespace of the cluster at the provided
+// context. If context is nil, the current context is used.
+func server(context *string) (uri, namespace string, err error) {
```
hausdorff: One last thing. `server` might not be the best name for this. Really you're resolving context to a URI and a namespace, right? Maybe `resolveContext`?

PR #28, opened 2017-10-23
 Set env namespace to default when namespace not provided

### LINKED ISSUE #3: Handle environment namespace defaults
_From @jessicayuen on October 6, 2017 21:2_

> Sorry, I forget what we decided here. When a user runs something like env set foo --namespace="", we have the following options:
> 
> Unset the namespace.
> Do nothing.
> Set namespace to "default".
> I'm fine pushing this to an issue and punting for now; I just want to make sure we make this decision purposefully. So why don't we do that?
> 
> That said, my inclination is to set it to "default". We also might also consider having an env unset command.

Also, consider removing `len(namespace) != 0` when overriding namespace, depending on the outcome of the decision above.

_Copied from original issue: ksonnet/kubecfg#172_

jessicayuen:
Wait for #24 to merge before review.

Fixes #3

`ks env set foo --namespace=""` and `ks env add foo` will now both set the
namespace to 'default' in the environment's spec.json file.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -31,6 +31,8 @@ const (
 	flagEnvURI       = "uri"
 	flagEnvNamespace = "namespace"
 	flagEnvContext   = "context"
+
+	defaultNamespace = "default"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


jessicayuen:
Yup sgtm, mentioned in PR body.

PR #35, opened 2017-10-26
Transition to pure Go implementation of Jsonnet

hausdorff:


utils/nativefuncs_test.go
```diff
@@ -60,22 +58,21 @@ func TestParseYaml(t *testing.T) {
...
     local a = std.native("parseYaml")("foo:\n- 3\n- 4\n")[0];
     a.foo[0] + a.foo[1]`)
-	check(t, err, x, "7\n")
+	check(t, err, x, "7")
 
 	x, err = vm.EvaluateSnippet("test", `
     local a = std.native("parseYaml")("---\nhello\n---\nworld");
     a[0] + a[1]`)
-	check(t, err, x, "\"helloworld\"\n")
+	check(t, err, x, "\"helloworld\"")
```
jessicayuen: I'm unsure where these native functions are called, but is removing the newline expected behavior?

hausdorff: Yep.

PR #43, opened 2017-11-03
Add rudimentary package management

hausdorff:


cmd/dep.go
```diff
@@ -0,0 +1,172 @@
...
+	depAddCmd.PersistentFlags().String(flagName, "", "Name to give the dependency")
+}
+
+var depCmd = &cobra.Command{
+	Use:   "dep",
+	Short: `Manage dependencies for the current ksonnet project`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return fmt.Errorf("Command 'dep' requires a subcommand\n\n%s", cmd.UsageString())
+	},
+	Long: ``,
```
jessicayuen: Delete this line and the below since not being used. Though, would be nice to see some a long description.

cmd/dep.go
```diff
@@ -0,0 +1,172 @@
...
+			return err
+		}
+		wd := metadata.AbsPath(cwd)
+
+		manager, err := metadata.Find(wd)
+		if err != nil {
+			return err
+		}
+
+		_, err = manager.CacheDependency(registry, libID, name, version)
```
jessicayuen: Did you want to move the non-command parsing logic over to the `/kubecfg/` package? Same applies for the other commands.

cmd/dep.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		for name := range app.Registries {
+			reg, _, err := manager.GetRegistry(name)
+			if err != nil {
+				return err
+			}
+
+			for libName := range reg.Libraries {
+				fmt.Println(libName)
```
jessicayuen: It would be helpful to also print the registry because:
1. it'll handle the event of shared libNames across registries
2. `dep add` requires the registry alongside the library name. would be helpful to the user when trying to add the dependency listed without also having to know of `registry list`

It's fine if you want to punt this until we generalize the pretty print though.

cmd/registry.go
```diff
@@ -0,0 +1,146 @@
...
+
+var regCmd = &cobra.Command{
+	Use:   "registry",
+	Short: `Manage registries for the current ksonnet project`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return fmt.Errorf("Command 'registry' requires a subcommand\n\n%s", cmd.UsageString())
+	},
+}
+
+var regAddCmd = &cobra.Command{
```
jessicayuen: Getting this unclear error:
```
➜  pr43 git:(pr-43) ✗ ../ks registry add blah
ERROR flag accessed but not defined: name
```

cmd/dep.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil
+	},
+	Long: `Cache a ksonnet library locally, and make it available for use in the current
+ksonnet project. This particularly means that we record enough information in
+'app.yaml' for new users to retrieve the dependency after a fresh clone of the
+app repository.
+
+For example, inside a ksonnet application directory, run:
+
+  ks dep get incubator/nginx@latest
```
jessicayuen: latest isn't supported, may want to update this.

metadata/registry_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+	}
+
+	for _, success := range successes {
+		success.Test(t)
+	}
+}
+
+//
+// TODO: Add failure tests, like:
```
jessicayuen: Might want to delete the below commented out tests before submitting, or add the tests.

cmd/registry.go
abiogenesis-now: ^ same

metadata/registry.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	// Check local disk cache.
+	gh, err := makeGitHubRegistryManager(registryRefSpec)
+	if err != nil {
+		return nil, err
+	}
+	registrySpecFile := string(m.registryPath(gh))
+
+	exists, _ := afero.Exists(m.appFS, registrySpecFile)
```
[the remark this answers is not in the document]
hausdorff: It's redundant. On error `exists` is `false`. But, we can do it here to report the error.

cmd/registry.go
hausdorff: Let's remove this command since we don't need it for this afternoon.

cmd/dep.go
hausdorff: Let's generalize later.

cmd/dep.go
hausdorff: Let's defer to after presentation.

hausdorff: https://github.com/ksonnet/ksonnet/issues/65

cmd/dep.go
hausdorff: https://github.com/ksonnet/ksonnet/issues/67

PR #23, opened 2017-10-19
[docs] Add autogenerated CLI reference

abiogenesis-now:
@jessicayuen @hausdorff PTAL

Copied from https://github.com/ksonnet/kubecfg/pull/130:

> I figured it might be useful to have autogenerated markdown docs for the CLI that users can peruse in addition to using --help in the CLI itself. Let me know if you think this is unnecessary though.

(`./docs/cli-reference` can also be incorporated as a Git submodule into the ksonnet website repo---granted with a lot of formatting changes :P) 

We should edit the actual CLI command descriptions at some point, but I'm holding off on that for now since some of them are still WIP.

PR #47, opened 2017-11-06
Add alias 'ks generate' for 'ks prototype use'

### LINKED ISSUE #46: `ks prototype use` giving wrong usage message
Note the second line says `preview`:

```
➜  ksonnet git:(master) ./ks prototype use
ERROR Command 'prototype preview' requires a prototype name

Usage:
  ks prototype use <prototype-name> <componentName> [type] [parameter-flags] [flags]

...
```

jessicayuen:
Also fixes #46 .

You will note that in the code change, I removed the command name from the error messages. This is because of
1. Shared logic between the two command names
2. Since we're printing the usage, I didn't find having the command name in the error message itself providing any additional help.

### DIFF
--- a/cmd/prototype.go
+++ b/cmd/prototype.go
@@ -30,6 +30,7 @@ import (
 
 func init() {
 	RootCmd.AddCommand(prototypeCmd)
+	RootCmd.AddCommand(generateCmd)
 	prototypeCmd.AddCommand(prototypeListCmd)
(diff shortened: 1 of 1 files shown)


abiogenesis-now:
^ after the commands are more solid (probably will be mostly there by thurs), I'll update and merge https://github.com/ksonnet/ksonnet/pull/23, and then do an actual review/edit of CLI content

PR #52, opened 2017-11-06
Jsonnet handling of component names with special characters

### LINKED ISSUE #51: Support hyphenated component names
The Jsonnet language does not support object identifiers with hyphens.

Ex: If a component is called `guestbook-ui`, we internally generate "component imports" in the form:

`guestbook-ui: import "/Users/jyuen/go/src/github.com/ksonnet/ksonnet/guestbook/components/guestbook-ui.jsonnet",`

Because the hyphen causes the Jsonnet to not compile, we will get the following error on apply:
```
➜  guestbook git:(ks-generate) ../ks apply default
ERROR Error reading /Users/jyuen/go/src/github.com/ksonnet/ksonnet/guestbook/environments/default/default.jsonnet: <extvar:__ksonnet/components>:2:12-13 Expected one of :, ::, :::, +:, +::, +:::, got: -

  guestbook-ui: import "/Users/jyuen/go/src/github.com/ksonnet/ksonnet/guestbook/components/guestbook-ui.jsonnet",
```

jessicayuen:
Currently, if a component name contains a special character, ex:
foo-bar, this translates to the jsonnet identifier: foo-bar, which is
invalid syntax.

This change will quote component names where there are special
characters.

Fixes #51

### DIFF
--- a/cmd/prototype.go
+++ b/cmd/prototype.go
@@ -25,6 +25,7 @@ import (
 	"github.com/ksonnet/ksonnet/metadata"
 	"github.com/ksonnet/ksonnet/prototype"
 	"github.com/ksonnet/ksonnet/prototype/snippet/jsonnet"
+	"github.com/ksonnet/ksonnet/utils"
 	"github.com/spf13/cobra"
 )
 
@@ -397,7 +398,11 @@ func expandPrototype(proto *prototype.SpecificationSchema, templateType prototyp
 		return "", err
 	}
 	if templateType == prototype.Jsonnet {
-		template = append([]string{`local params = std.extVar("` + metadata.ParamsExtCodeKey + `").components.` + componentName + ";"}, template...)
+		componentsText := "components." + componentName
+		if !utils.IsASCIIIdentifier(componentName) {
+			componentsText = fmt.Sprintf(`components["%s"]`, componentName)
+		}
+		template = append([]string{`local params = std.extVar("` + metadata.ParamsExtCodeKey + `").` + componentsText + ";"}, template...)
 	}
 
 	return jsonnet.Parse(componentName, strings.Join(template, "\n"))
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -27,6 +27,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/ksonnet/ksonnet/metadata/params"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	log "github.com/sirupsen/logrus"
@@ -430,6 +432,7 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
 //
 //   {
 //      foo: import "components/foo.jsonnet"
+//      "foo-bar": import "components/foo-bar.jsonnet"
 //   }
 func constructBaseObj(paths []string) string {
 	var obj bytes.Buffer
@@ -441,6 +444,7 @@ func constructBaseObj(paths []string) string {
 		}
 
 		name := strings.TrimSuffix(path.Base(p), ext)
+		name = params.SanitizeComponent(name)
 		fmt.Fprintf(&obj, "  %s: import \"%s\",\n", name, p)
 	}
 	obj.WriteString("}\n")
--- a/metadata/params/params.go
+++ b/metadata/params/params.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/google/go-jsonnet/ast"
 	"github.com/google/go-jsonnet/parser"
+	"github.com/ksonnet/ksonnet/utils"
 )
 
 const (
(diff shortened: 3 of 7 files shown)


cmd/prototype.go
```diff
@@ -397,7 +398,11 @@ func expandPrototype(proto *prototype.SpecificationSchema, templateType prototyp
 		return "", err
 	}
 	if templateType == prototype.Jsonnet {
-		template = append([]string{`local params = std.extVar("` + metadata.ParamsExtCodeKey + `").components.` + componentName + ";"}, template...)
+		componentsText := "components." + componentName
+		if utils.HasSpecialCharacter(componentName) {
+			componentsText = fmt.Sprintf(`components["%s"]`, componentName)
+		}
+		template = append([]string{`local params = std.extVar("` + metadata.ParamsExtCodeKey + `").` + componentsText + ";"}, template...)
```
hausdorff: We could just simplify this by using `components["whatever"]` syntax for _every_ component, no? The way we're doing it now is cleaner, so i think it's nbd either way, just thought I'd bring up the alternatives and see what you think.

utils/strings.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"strings"
+)
+
+// HasSpecialCharacter takes a string and returns true if the string contains
+// any special characters.
+func HasSpecialCharacter(s string) bool {
+	f := func(r rune) bool {
+		return r < 'A' || r > 'z'
```
hausdorff: So, formally, Jsonnet can't lex unicode identifiers. So this logic is probably what we want. But, I think `HasSpecialCharacters` is a little opaque. I would call it `IsASCIIIdentifier` and invert the logic. Because that's what we want: if `!IsASCIIIdentifier` then quote.

utils/strings_test.go
```diff
@@ -0,0 +1,49 @@
...
+			expected: true,
+		},
+		{
+			input:    "helloworld",
+			expected: false,
+		},
+		{
+			input:    "hello-world",
+			expected: true,
+		},
```
hausdorff: Might want to toss in a non-ASCII unicode character here too just for fun.

metadata/params/params.go
```diff
@@ -39,6 +40,30 @@ func astRoot(component, snippet string) (ast.Node, error) {
...
+	if field.Id != nil && string(*field.Id) == component {
+		return true
+	}
+	// case "foo-bar": {...}
+	if field.Expr1 != nil {
+		c, isStringLiteral := field.Expr1.(*ast.LiteralString)
+		if isStringLiteral && c.Value == component {
+			return true
+		}
+	}
```
hausdorff: The `ast.ObjectField` type has a member, `Kind`, which you can switch on here to determine whether `Id` is populated or not. You should need to handle only the `ObjectFieldID` and `ObjectFieldStr` cases:

```
switch field.Kind {
case ObjectFieldStr: [... expr1 logic here ...]
case ObjectFieldID: [... id logic here ...]
}
```

hausdorff: Also this will break if the literal string is a block. We should probably switch and emit the correct string type just in case.

metadata/params/params.go
```diff
@@ -71,7 +96,18 @@ func visitAllParams(components ast.Object) (map[string]Params, error) {
...
+			id = string(*f.Id)
+		} else {
+			c, isStringLiteral := f.Expr1.(*ast.LiteralString)
+			if isStringLiteral {
+				id = c.Value
+			} else {
+				return nil, fmt.Errorf("Found an unsupported component type %T", f.Expr1)
+			}
+		}
+		params[id] = p
```
hausdorff: I think my comments about the object fields above are probably relevant here, too.

cmd/prototype.go
jessicayuen: Think I'm going to leave this since `components.whatever` is cleaner. Prefer to use that syntax where we can.

metadata/params/params.go
```diff
@@ -71,7 +105,22 @@ func visitAllParams(components ast.Object) (map[string]Params, error) {
...
 		}
-		params[string(*f.Id)] = p
+		var id string
+		switch f.Kind {
+		case ast.ObjectFieldStr:
+			c, _ := f.Expr1.(*ast.LiteralString)
+			switch c.Kind {
+			case ast.StringSingle:
+				fallthrough
+			case ast.StringDouble:
```
hausdorff: Mentioned offline, but plan is to use multi-case syntax, like: `case ast.StringSingle, ast.StringDouble:`. `fallthrough` is fraught.

cmd/root.go
```diff
@@ -441,6 +442,9 @@ func constructBaseObj(paths []string) string {
 		}
 
 		name := strings.TrimSuffix(path.Base(p), ext)
+		if !utils.IsASCIIIdentifier(name) {
+			name = fmt.Sprintf(`"%s"`, name)
```
hausdorff: Tiny nit, isn't this just `sanitizeComponent`?

metadata/params/params.go
```diff
@@ -71,7 +105,22 @@ func visitAllParams(components ast.Object) (map[string]Params, error) {
 		if err != nil {
 			return nil, err
 		}
-		params[string(*f.Id)] = p
+		var id string
+		switch f.Kind {
+		case ast.ObjectFieldStr:
```
hausdorff: Am I missing something, or does it make sense to just have `hasComponent` return a string as well? This seems very similar?

utils/strings_test.go
```diff
@@ -0,0 +1,53 @@
...
+		{
+			input:    "helloworld",
+			expected: true,
+		},
+		{
+			input:    "hello-world",
+			expected: false,
+		},
+		{
+			input:    "helloworld&",
```
hausdorff: lol `&` is not a non-ASCII unicode character. ;) but whatever.

cmd/root.go
jessicayuen: yes, but `sanitizeComponent` isn't visible outside of params package ... do you think it should be public?

hausdorff: I'd make this public, just so the name munging logic is in one place. Then come back and clean this up when we move `constructBaseObj` somewhere else.

PR #58, opened 2017-11-07
Renaming environment should also rename nested .jsonnet

### LINKED ISSUE #54: Renaming env needs to also rename the nested jsonnet file
In the demo, doing
```
ks env set default --name=dev
```
renames `environments/default` to `environments/dev` *but* fails to rename `environments/default/default.jsonnet` to `environments/dev/dev.jsonnet`

jessicayuen:
Currently the env directory is being renamed on `env set foo
--name=bar`, however the `foo.jsonnet` file also needs to be renamed to
`bar.jsonnet`. This commit fixes that.

Fixes #54

### DIFF
--- a/metadata/environment.go
+++ b/metadata/environment.go
@@ -180,23 +180,9 @@ func (m *manager) DeleteEnvironment(name string) error {
 
 	// Need to ensure empty parent directories are also removed.
 	log.Debug("Removing empty parent directories, if any")
-	parentDir := name
-	for parentDir != "." {
-		parentDir = filepath.Dir(parentDir)
-		parentPath := string(appendToAbsPath(m.environmentsPath, parentDir))
-
-		isEmpty, err := afero.IsEmpty(m.appFS, parentPath)
-		if err != nil {
-			log.Debugf("Failed to check whether parent directory at path '%s' is empty", parentPath)
-			return err
-		}
-		if isEmpty {
-			log.Debugf("Failed to remove parent directory at path '%s'", parentPath)
-			err := m.appFS.RemoveAll(parentPath)
-			if err != nil {
-				return err
-			}
-		}
+	err = m.cleanEmptyParentDirs(name)
+	if err != nil {
+		return err
 	}
 
 	log.Infof("Successfully removed environment '%s'", name)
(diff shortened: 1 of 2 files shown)


metadata/environment.go
```diff
@@ -293,12 +293,20 @@ func (m *manager) SetEnvironment(name string, desired *Environment) error {
...
-		pathOld := string(appendToAbsPath(m.environmentsPath, name))
-		pathNew := string(appendToAbsPath(m.environmentsPath, desired.Name))
-		log.Debugf("Moving directory at path '%s' to '%s'", pathOld, pathNew)
-		err = m.appFS.Rename(pathOld, pathNew)
+		pathOld := appendToAbsPath(m.environmentsPath, name)
+		pathNew := appendToAbsPath(m.environmentsPath, desired.Name)
+		jsonnetPathOld := appendToAbsPath(pathOld, path.Base(name)+".jsonnet")
+		jsonnetPathNew := appendToAbsPath(pathOld, desired.Name+".jsonnet")
+		log.Debugf("Renaming jsonnet file from '%s' to '%s'", path.Base(string(jsonnetPathOld)), path.Base(string(jsonnetPathNew)))
+		err = m.appFS.Rename(string(jsonnetPathOld), string(jsonnetPathNew))
```
hausdorff: I think this isn't going to work unless you create the new directory first. Right? So renaming say `dev` to `us-west/dev` will fail. This is tricky so we probably want to add tests, also.

jessicayuen: No, this will work. afero's `rename` is a bit misleading that it actually does the equivalent of `mv`.

There's also an open issue with afero that isn't able to allow me to add the unit test I want. (Because testFS won't move the file contents despite the actual code working): https://github.com/spf13/afero/issues/141

hausdorff: Sorry, what I'm saying is that renaming some environment `bar` -> `us-west/bar` will fail because we're moving the `.jsonnet` file to the `us-west/` directory before creating it, which will fail.

I've tested and confirmed:

```
➜  simple git:(pr-58) ✗ ../ks env set bar --name us-west/dev
INFO  Setting environment name from 'bar' to 'us-west/dev'
ERROR rename /Users/alex/src/go/src/github.com/ksonnet/ksonnet/simple/environments/bar/bar.jsonnet /Users/alex/src/go/src/github.com/ksonnet/ksonnet/simple/environments/bar/us-west/dev.jsonnet: no such file or directory
```

hausdorff:
So, as I mentioned offline, the current regimen will "orphan" directories. If you rename `us-west/prod` to `prod` it will leave an empty `us-west` lying around. We should fix that before we ship.

PR #60, opened 2017-11-07
Pass context & overrides by reference when initiating client

### LINKED ISSUE #53: `ks apply <env>` isn't being deployed to the correct context
It works if the `--context` flag is supplied, but otherwise does not deploy to the context / namespace specified by the environment.

jessicayuen:
This fixes the bug where objects weren't being deployed to the cluster
specified by the environment. This bug occurred because overrides (ex:
cluster, namespace) were being lost because we weren't applying the
changes to the correct override object. To fix this, we needed to pass
the overrides by reference.

Fixes #53


jessicayuen:
Fixed!

jessicayuen:
Actually -- not fixed, accidentally merged some sames from another branch. Hold off...

jessicayuen:
Actually -- not fixed, accidentally merged some changes from another branch. Hold off...

jessicayuen:
ok, good call. Can you take a look again?

PR #63, opened 2017-11-07
Add ability to retrieve prototypes from vendor/

hausdorff:
Just a note, the prototype comment parsing comes out of the ksonnet/parts doc gen code, so this is not actually new. We will eventually pull that out and vendor this into that project.

### DIFF
--- a/cmd/prototype.go
+++ b/cmd/prototype.go
@@ -92,7 +92,23 @@ var prototypeListCmd = &cobra.Command{
 			return fmt.Errorf("Command 'prototype list' does not take any arguments")
 		}
 
-		index := prototype.NewIndex([]*prototype.SpecificationSchema{})
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		wd := metadata.AbsPath(cwd)
+
+		manager, err := metadata.Find(wd)
+		if err != nil {
+			return err
+		}
+
+		extProtos, err := manager.GetAllPrototypes()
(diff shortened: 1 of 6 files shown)


cmd/prototype.go
```diff
@@ -92,7 +92,23 @@ var prototypeListCmd = &cobra.Command{
...
 		}
 
-		index := prototype.NewIndex([]*prototype.SpecificationSchema{})
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		wd := metadata.AbsPath(cwd)
+
+		manager, err := metadata.Find(wd)
```
jessicayuen: We should probably open a ticket to separate the prototype logic into the `kubecfg` package so we don't have to have the redundant code in getting `manager`.

hausdorff: Prototype code probably can't go into the kubecfg package, since that's going to be a shared package. But, can you tell me more about which code you see as being rendundant?

jessicayuen: Hm, I've always thought of `kubecfg` as a package for ksonnet that handles more internal command logic outside of command flag parsing done by `cmd`.

I'm referring to the common code that gets the working directory, and then finds the manager. I added a shared function `manager()` in the `kubecfg` package that does that in an earlier commit.

hausdorff: Ok, we can do that. https://github.com/ksonnet/ksonnet/issues/65

Issue #68, opened 2017-11-08
Add dependencies to -J path


abiogenesis-now:
@hausdorff @jessicayuen 

- **Vendor path**
  `ks apply` and `ks default` don't work for components generated from the `incubator` prototypes--*this is because `vendor/` is not included in the default jsonnet lib paths.* I'm not sure if `root.go` needs to be modified to include it or somewhere else; either way, I don't think it'll be a difficult fix.
  - [x] Adding `vendor/` to the `-J` paths in ksonnet/ksonnet.
  - [ ] Flatten out the `vendor/` directory.
  - [ ] Changing all prototypes in ksonnet/parts from `import "incubator/whatever/whatever.libsonnet"` to use `import "whatever/whatever.libsonnet"` instead.

hausdorff:
@jessicayuen @abiogenesis-now I'd like to move this to two different issues. Since you're in an urgent meeting I hope you don't mind if I just do it.

abiogenesis-now:
yah that's fine, feel free to close this or modify however you please

hausdorff:
Alright, @jessicayuen @abiogenesis-now let's add `vendor/` to the ksonnet path. This will involve two steps:

1. Adding `vendor/` to the `-J` paths in ksonnet/ksonnet
1. Changing all prototypes to `import "incubator/whatever/whatever.libsonnet"` to use `import "whatever/whatever.libsonnet"` instead.

I'll update the issue. If something sounds wrong we can re-work the issue again.

abiogenesis-now:
@hausdorff wait I think it works fine if prototypes use `import "incubator/whatever/whatever.libsonnet"` in their definitions? e.g. because of the directory structure they need that registry prefix

hausdorff:
One more thing. Right now `vendor/` has the registry name baked into the file structure. So imports are of the form `vendor/incubator/nginx`. This is a mistake. It complicates everything (_e.g._, renaming registries) and is not materially useful for the dependencies themselves.

So, we need to flatten it out, THEN we can do the second step.

I'll update the issue.

hausdorff:
Bumping down to p1. Flattening the `vendor/` directory is not our highest priority right now.

hausdorff:
Bumping to P1, since #262 is P1, and that's a duplicate of this one.

bryanl:
ksonnet adds vendor to the jsonnet path.

PR #71, opened 2017-11-08
Add `vendor/` to Jsonnet paths

hausdorff:
This resolves the first half of #68. When we `generate` using a
prototype from a vendored dependency, it often results in a compilation
error when we `apply`, because these prototypes usually depend on code
that exists in the dependency, and `vendor/` is not a part of the
Jsonnet search path.

This commit resolves this problem by adding it to the search path.

### DIFF
--- a/metadata/manager_test.go
+++ b/metadata/manager_test.go
@@ -263,15 +263,19 @@ func TestComponentPaths(t *testing.T) {
 
 func TestLibPaths(t *testing.T) {
 	appName := "test-lib-paths"
+	expectedVendorPath := path.Join(appName, vendorDir)
 	expectedLibPath := path.Join(appName, libDir)
 	expectedEnvLibPath := path.Join(appName, environmentsDir, mockEnvName, metadataDirName)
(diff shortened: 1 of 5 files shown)


metadata/manager_test.go
```diff
@@ -263,16 +263,20 @@ func TestComponentPaths(t *testing.T) {
...
 	expectedEnvParamsPath := path.Join(appName, environmentsDir, mockEnvName, paramsFileName)
 	m := mockEnvironments(t, appName)
 
-	libPath, envLibPath, envComponentPath, envParamsPath := m.LibPaths(mockEnvName)
+	libPath, vendorPath, envLibPath, envComponentPath, envParamsPath := m.LibPaths(mockEnvName)
 	if string(libPath) != expectedLibPath {
 		t.Fatalf("Expected lib path to be:\n  '%s'\n, got:\n  '%s'", expectedLibPath, libPath)
 	}
+	if string(vendorPath) != expectedVendorPath {
+		t.Fatalf("Expected lib path to be:\n  '%s'\n, got:\n  '%s'", expectedLibPath, libPath)
```
jessicayuen: this should be `expectedVendorPath` and `vendorPath`

hausdorff: Yeah, good catch.

PR #73, opened 2017-11-08
Add optional params to prototype spec

hausdorff:
Currently we allow users to specify prototypes using Jsonnet. At the
head of each prototype should be a comment with semi-structured
metadata, e.g.,

```
  // @apiVersion 0.0.1
  // @name io.ksonnet.pkg.nginx-simple
  // @description Deploys a simple, stateless nginx server with server
  // blocks (roughly equivalent
  //   to nginx virtual hosts). The nginx container is deployed using a
  //   Kubernetes deployment, and is exposed to a network with a service.
  // @param namespace string default Namespace in which to put the
  // application
  // @param name string Name to give to each of the components
```

Critically, some metadata about params is stored using the `@param` tag,
including name, type, and description, for pretty printing.

This commit will introduce a new param type, `@optionalParam`. This type
has a name, a type, and a default value.

It is currently not possible to specify a default value with spaces, but
this should be good enough for MVP.

### DIFF
--- a/prototype/specification.go
+++ b/prototype/specification.go
@@ -19,6 +19,7 @@ const (
 	nameTag        = "@name"
 	descriptionTag = "@description"
 	paramTag       = "@param"
+	optParamTag    = "@optionalParam"
 )
 
 func FromJsonnet(data string) (*SpecificationSchema, error) {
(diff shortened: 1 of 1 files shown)


prototype/specification.go
```diff
@@ -222,6 +223,27 @@ func (s *SpecificationSchema) addField(tag, text string) error {
 			Default:     nil,
 			Type:        pt,
 		})
+	case optParamTag:
+		// NOTE: There is usually more than one `@param`, so we don't
```
jessicayuen: s/@param/@optionalParam

PR #74, opened 2017-11-08
De/serialize specification objects with YAML

### LINKED ISSUE #48: Make sure dep management supports YAML metadata files as well as JSON
Marshaling in https://github.com/hausdorff/ksonnet/blob/395491ab0aeb6d853e235aa1a4025ee7a14b8c76/metadata/app/schema.go should support YAML.

hausdorff:
Fixes #48.

Currently the structs `app.Spec`, `registry.Spec`, and `parts.Spec` are
all serialized and deserialized using JSON instead of YAML.

This commit will resolve this issue.

### DIFF
--- a/metadata/registry.go
+++ b/metadata/registry.go
@@ -1,11 +1,11 @@
 package metadata
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	"github.com/ghodss/yaml"
 	"github.com/ksonnet/ksonnet/metadata/app"
 	"github.com/ksonnet/ksonnet/metadata/parts"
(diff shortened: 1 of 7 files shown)


metadata/registry.go
```diff
@@ -1,11 +1,12 @@
 package metadata
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	yaml "gopkg.in/yaml.v2"
```
jessicayuen: Any reason we're using this package here instead of using what the other places are using: `"github.com/ghodss/yaml"`?

hausdorff: Arg.

PR #79, opened 2017-11-09
💅 Add row-padding string pretty printer

hausdorff:
Historically, every time we've needed to write out something tabular,
we've hand-rolled an ad hoc padded row writer. Yes, in each case. Such
is #startuplife.

The third time we did this we wrote this with bugs, and so now we are
rewriting it in a general, tested function, and transitioning all places
where we call this to use this function.

### DIFF
--- a/pkg/kubecfg/param.go
+++ b/pkg/kubecfg/param.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	param "github.com/ksonnet/ksonnet/metadata/params"
+	"github.com/ksonnet/ksonnet/utils"
 
 	"github.com/fatih/color"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


hausdorff:
@jessicayuen Nope. It's because I thought you colorized `param list` but maybe not.

jessicayuen:
> @jessicayuen Nope. It's because I thought you colorized param list but maybe not.

Nope, just colorized `param diff`.

pkg/kubecfg/param.go
```diff
@@ -193,53 +194,23 @@ func outputParamsFor(component string, params param.Params, out io.Writer) error
...
-	//   foo       replicas  1
-	//
-
-	maxComponentLen := len(paramComponentHeader)
-	for _, k := range keys {
-		if l := len(k); l > maxComponentLen {
-			maxComponentLen = l
-		}
+	rows := [][]string{
+		[]string{"COMPONENT", "PARAM", "VALUE"},
```
jessicayuen: There's constants on line 110 `paramComponentHeader`, etc. Can you use those instead?

PR #87, opened 2017-11-13
Add more useful error when attempting to 'show' files in components/

### LINKED ISSUE #13: Better error messaging when env is not supplied for `kubecfg show -f <FILE>`
_From @abiogenesis-now on September 11, 2017 21:15_

If an `env` isn't supplied with the `kubecfg show -f <FILE>` command, `kubecfg` can't load the appropriate ksonnet lib files from `/environments` with the `-J` flag.

We need better error messaging so the user knows what to do in this scenario.

_Copied from original issue: ksonnet/kubecfg#127_

jessicayuen:
If an env isn't supplied with the kubecfg show -f <FILE> command, where
the file is present in the components/ directory, kubecfg can't load the
appropriate ksonnet lib files from /environments with the -J flag.

This commit adds an INFO message on `show` failures related to expanding
the jsonnet file that provides the user with more context.

Fixes #13

### DIFF
--- a/cmd/show.go
+++ b/cmd/show.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/ksonnet/ksonnet/metadata"
@@ -67,6 +68,8 @@ var showCmd = &cobra.Command{
 
 		objs, err := expandEnvCmdObjs(cmd, envSpec, wd)
 		if err != nil {
+			log.Info("Jsonnet files in components/ aren't able to be expanded without an environment. " +
+				"Consider running this command with the format: `ks show <env> -f <file>.")
 			return err
 		}
 


cmd/show.go
```diff
@@ -66,10 +67,13 @@ var showCmd = &cobra.Command{
 		}
 
 		objs, err := expandEnvCmdObjs(cmd, envSpec, wd)
+		out := cmd.OutOrStdout()
```
hausdorff: Does this need to be pulled into an a variable? I don't think we're using it outside of `c.Run`. (Also it seems especially odd to put it between the declaration of `err` and `if err != nil`.)

jessicayuen:
I don't think we can make the show command require an env arg for reasons of wanting to show files outside of the ksonnet project structure. Ex: `ks show -f examples/guestbook.jsonnet`.

cmd/show.go
jessicayuen: whoops ... I was wanting to use `out` for the outputting earlier instead of `log` but forgot to change back ..

hausdorff:
@jessicayuen I guess what I'm saying is that maybe we _don't_ want to allow this at this point. I don't think people will really care that much about stuff outside of the project.

jessicayuen:
@hausdorff Hmm I don't think we should take it away for the same reasons that we currently allow `ks apply -f examples/guestbook.jsonnet`.

jessicayuen:
@hausdorff Opened #100 from our offline discussions.

jessicayuen:
We are going to remove the standalone `-f` flag, see #100

PR #91, opened 2017-11-14
Implement `registry describe` and `dep describe`

hausdorff:


cmd/registry.go
```diff
@@ -0,0 +1,136 @@
...
+	registryCmd.AddCommand(registryListCmd)
+	registryCmd.AddCommand(registryDescribeCmd)
+}
+
+var registryCmd = &cobra.Command{
+	Use:   "registry",
+	Short: `Manage registries for current project`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return fmt.Errorf("Command 'registry' requires a subcommand\n\n%s", cmd.UsageString())
+	},
```
jessicayuen: Can you add a long description here about what registries are?

cmd/registry.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		formatted, err := utils.PadRows(rows)
+		if err != nil {
+			return err
+		}
+		fmt.Print(formatted)
+		return nil
+	},
+	Long: ``,
```
jessicayuen: Delete the `Long` and `Example` if not going to fill in. Same with Describe below.

cmd/registry.go
```diff
@@ -0,0 +1,136 @@
...
+
+	Example: ``,
+}
+
+var registryDescribeCmd = &cobra.Command{
+	Use:   "describe",
+	Short: `Manage registries for current project`,
```
jessicayuen: Shouldn't this be `Describe registries....`?

cmd/dep.go
```diff
@@ -102,6 +106,65 @@ With these commands, a user can 'import "kspkg://nginx1"', and
 'import "kspkg://nginx2"' with no conflict.`,
 }
 
+var depDescribeCmd = &cobra.Command{
+	Use:   "describe [<registry-name/]<package-name>",
```
jessicayuen: `/` should be `>` in `<registry-name/`

hausdorff: I'm changing it to `[<registry-name>/]`, because the `/` is not optional if you include `registry-name`.

PR #97, opened 2017-11-15
Normalize environment server URL

jessicayuen:
As part of the environments feature, we want to be able to deploy to a
specific cluster given the environment's URI. We cross-check against
kubecfg for this URI's location. For fail-safe comparison, we probably
want to normalize these URIs.


jessicayuen:
@hausdorff Strangely, the dependency is being used somewhere already!

PR #98, opened 2017-11-15
Rename <env>.jsonnet file to main.jsonnet

### LINKED ISSUE #76: Support renaming of environments to parent directories
Ex: `ks env set us-west/prod --name=us-west`. Currently erroring out.

### LINKED ISSUE #77: Rename all `<env>.jsonnet` files to `main.jsonnet`
This will make the logic around renaming environments and referring to this file simpler if it is consistent across all environments.

jessicayuen:
Currently, creating a `dev` environment will create a file
`dev.jsonnet`. Creating a `prod` environment will create a file
`prod.jsonnet`. This is a little more complex and prone to error than it
needs to be, especially when we are renaming environments. It will also
make this file easier to refer to in documentation if we give the file a
static name -- `main.jsonnet`.

Fixes #77 

--

Also, support renaming of environments to parent and child directories.

Fixes #76

### DIFF
--- a/metadata/environment.go
+++ b/metadata/environment.go
@@ -47,6 +47,17 @@ const (
 	specFilename   = "spec.json"
 )
 
+var envPaths = []string{
+	// metadata Dir.wh
+	metadataDirName,
+	// environment base override file
+	envFileName,
+	// params file
+	paramsFileName,
+	// spec file
+	specFilename,
+}
+
 // Environment represents all fields of a ksonnet environment
 type Environment struct {
 	Path      string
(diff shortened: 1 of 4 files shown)


jessicayuen:
@hausdorff fixed!

metadata/environment.go
```diff
@@ -281,37 +285,43 @@ func (m *manager) SetEnvironment(name string, desired *Environment) error {
...
+			m.mvEnvFileContents(pathOld, pathNew)
+		} else if filepath.HasPrefix(string(pathNew), string(pathOld)) {
+			// the new directory is a children of the old directory --
+			// rename won't work.
+			err = m.appFS.MkdirAll(string(pathNew), defaultFolderPermissions)
+			if err != nil {
+				return err
+			}
+			m.mvEnvFileContents(pathOld, pathNew)
+		} else {
```
hausdorff: A minor thing we should know about going in is that this allows us to have both an environment `us-west` and `us-west/dev`. I think this kinda sucks, but I don't feel super strongly about it.

jessicayuen: It's intentional -- agree it might not be *great* choice but I think this isn't a limitation we should impose on the user, if they really wanted to do that for some reason.

metadata/environment.go
```diff
@@ -418,6 +428,37 @@ func (m *manager) SetEnvironmentParams(env, component string, params param.Param
...
+		// params file
+		paramsFileName,
+		// spec file
+		specFilename,
+	}
+	for _, p := range basePaths {
+		if err := m.appFS.Rename(
+			string(appendToAbsPath(dirPathOld, p)),
+			string(appendToAbsPath(dirPathNew, p))); err != nil {
+			return err
```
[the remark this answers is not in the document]
jessicayuen: Agree with both points!

jessicayuen:
@hausdorff PTAL

abiogenesis-now:
yes glad to see this is happening!

metadata/environment.go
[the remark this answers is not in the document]
hausdorff: Bug filed. https://github.com/ksonnet/ksonnet/issues/136

metadata/environment.go
hausdorff: I thought the plan was to make this a hierarchy. Otherwise `environments` could just be a flat directory of directories like `us-west_dev`. Let's file a bug and discuss: https://github.com/ksonnet/ksonnet/issues/138

PR #104, opened 2017-11-17
[docs] Concepts Reference

abiogenesis-now:
@jessicayuen @hausdorff 

This PR adds a reference for all the ksonnet concepts.

A lot of this will probably be reused in the tutorial somehow, but I'm going to figure out the mechanics of that after the fact. For now, not going to worry about duplicate content.


abiogenesis-now:
@jessicayuen @hausdorff fyi, this is ready for review! PTAL, thank you :)

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+</p>
+
+All of the component files in an *app* can be deployed to a specified *environment* using [`ks apply`](docs/cli-reference/ks_apply.md).
+
+---
 
 ### Prototype
-TODO
+
+Prototypes are the basis for components—they are the reason that `ks generate` allows you to quickly move from a blank app to multiple complete manifests.
```
seh: - [ ] s/—/;/  
  If you don't like the semicolon, consider a colon instead.

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+
+By itself, a prototype is an pre-written but incomplete manifest:
```
seh: - [ ] s/an/a/

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+
+<p align="center">
+<img alt="prototype diagram" src="/docs/img/prototype.svg" height="170px">
+</p>
+
+Why is this useful? Prototypes allow you to avoid copying and pasting boilerplate code, and to focus on the parts of your configuration that are specific to your app. Specifically, during `ks generate`, you can specify certain command-line *parameters* to "fill-in-the-blank(s)" of a prototype and output a complete `component/` manifest:
+
+![prototype parameter component diagram](/docs/img/prototype_and_parameters.svg)
+
+Out of the box, ksonnet comes with some system prototypes (like `io.ksonnet.pkg.deployment-exposed-with-service`) that you can explore with the [`ks prototype`](docs/cli-reference/ks_prototype.md) commands. See [*package*](#package) for information on downloading or sharing additional prototypes.
```
seh: - [ ] s/commands/command/  
  Is _ks prototype_ multiple commands? The sentence reads strangely with "commands" pluralized.

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+
+Prototypes are the basis for components—they are the reason that `ks generate` allows you to quickly move from a blank app to multiple complete manifests.
+
+By itself, a prototype is an pre-written but incomplete manifest:
+
+<p align="center">
+<img alt="prototype diagram" src="/docs/img/prototype.svg" height="170px">
+</p>
+
+Why is this useful? Prototypes allow you to avoid copying and pasting boilerplate code, and to focus on the parts of your configuration that are specific to your app. Specifically, during `ks generate`, you can specify certain command-line *parameters* to "fill-in-the-blank(s)" of a prototype and output a complete `component/` manifest:
```
seh: I wrote a comment here, and lost it. I'll restate it succinctly.

- [ ] Write "fill-in-the-blanks" pluralized, as it's sufficiently unlikely that a prototype will fewer than two parameters.
- [ ] Is the trailing slash after "componet/" intentional?

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+  * [Package](#package)
+  * [Registry](#registry)
+* Related
+  * [Manifest](#manifest)
+  * [Jsonnet](#jsonnet)
+
+
+The following diagram shows how the ksonnet framework works holistically:
+
+![ksonnet overview diagram](/docs/img/ksonnet_overview.svg)
```
jessicayuen: Would it be possible to parallelize the horizontal arrows (---->) and make them the same width/distance? Initial look made me think that `component` was a part of  the first grouping.

I think it would also be helpful if all the components were the same shape as `component` (the cylinder).

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+
+---
 
 ### Application
-TODO
+
+A ksonnet application represents a well-structured directory of Kubernetes [manifests](#manifests). This directory is autogenerated by [`ksonnet init`](docs/cli-reference/ksonnet_init.md). These manifests typically tie together in some way—collectively, they might define something like the following:
```
jessicayuen: Would be helpful here I think to give some examples of what they define together, or call out what that diagram actually is.

`... like the following web service:`

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
 
 ### Environment
-TODO
+
+An environment consists of four elements:
+
+| What       | Example       | Source          | How is this tracked? |
+| ---------- | ------------- | ------------- | ---------- |
+| (1) *Name* | dev | Specified at command line | A directory under `environments/` (e.g. `environments/dev/`). <br><br>**A name with slashes results in a hierarchical file structure**. For example, `us-west/staging` creates `environments/us-west/staging/`.|
+| (2) *Cluster URI* | https://cluster-name-with-hash.us-west-2.elb.amazonaws.com | Pulled from current [kubeconfig context](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context) OR specified at command line | Inside `environments/<env-name>/spec.json` |
```
jessicayuen: s/Server/Cluster URI. (Since server is the terminology we changes for the command line because it more closely aligns with kubectl)

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+| (3) *Kubernetes [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/)* | dev | Pulled from current [kubeconfig context](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context) OR specified at command line | Inside `environments/<env-name>/spec.json` |
```
jessicayuen: I'm not sure if the `Source` column would be particularly useful. Because `namespace` and `server` can also be defined at the command line. I think the specifics of this are best left to the command line docs.

I think it might be more helpful to give a description of what each of these are.

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
 ### Component
-TODO
+
+A ksonnet *application* can be broken down into a series of discrete components:
+
+<p align="center">
+<img alt="ksonnet component diagram" src="/docs/img/component.svg" height="150px">
+</p>
+
+Components can be as simple as a single Kubernetes resource (e.g. a Deployment) or as complex as a complete logging stack (e.g. EFK). More concretely, a component corresponds to a Kubernetes manifest in `components/`. This manifest is autogenerated by the [`ks generate`](docs/cli-reference/ks_generate.md) command, and expressed in a language called [Jsonnet](#jsonnet).
```
jessicayuen: I don't think we should be tightly coupling components with prototypes through the generate command. I can easily see users wanting to add static manifests. Maybe `This manifest is typically autogenerated...`

docs/concepts.md
```diff
@@ -1,27 +1,213 @@
...
+
+How are these components generated? All components are generated from *prototypes*. The distinction between a component and a prototype is a bit subtle. If you are familiar with object oriented programming, a prototype is like a class, and a component is like its instantiation:
```
jessicayuen: Same as above; I wouldn't call out that all components are generated from prototypes. This might change the story slightly for *params*, as well.

abiogenesis-now: Fair point! I'll change that

abiogenesis-now:
note to self: fix picture

thanks @jessicayuen and @seh for the much-appreciated reviews!! Feel free to follow up with additional suggestions in an issue and tag with "documentation"; we're just merging this now so we can send the alpha release email.

PR #108, opened 2017-11-17
[docs] Fix ksonnet concept overview image + minor doc nits

abiogenesis-now:
This PR has some miscellaneous fixes:

* Updates the `ksonnet_overview.svg` diagram according to @jessicayuen 's feedback
* Emphasizes the point about migration in [*components*](/docs/concepts.md#components)
* Changes the README instructions to be more clear about checking your `$GOPATH`, based on feedback from @k8k (our target users are newish Kubernetes users, who may not be familiar with all the ins and outs of Go)


README.md
```diff
@@ -18,40 +18,51 @@ All of this results in a more iterative process for developing manifests, one th
...
-make install
+2. Now, to install ksonnet, run the following commands:
 
-# Ensure that Go binaries are runnable in any directory
-PATH=$PATH:$GOPATH/bin
-```
+    ```bash
+    # Make sure that your $GOPATH is set up! This is usually /usr/local/
+    # /usr
+    echo $GOPATH
```
jessicayuen: Think you can delete this line since it is the same as step 1.

README.md
```diff
@@ -18,40 +18,51 @@ All of this results in a more iterative process for developing manifests, one th
...
 
-If your ksonnet is properly installed, you should be able to run the following `--help` command and see similar output:
+    # Download ksonnet
+    go get github.com/ksonnet/ksonnet
 
-```
-$ ks --help
+    # Separate make command to install binary under shortname `ks`
```
jessicayuen: Do you think we can simplify this to 
```
Build and install binary under shortname `ks`
```

README.md
```diff
@@ -18,40 +18,51 @@ All of this results in a more iterative process for developing manifests, one th
...
+    cd $GOPATH/src/github.com/ksonnet/ksonnet
+    make install
 
-Synchronise Kubernetes resources with config files
+    # Ensure that Go binaries are runnable in any directory
```
jessicayuen: I would move this up to step 1, so we can distinguish the ksonnet specific install from the go install.

README.md
abiogenesis-now: oops, good catch

README.md
abiogenesis-now: yes, much better, thank you!

README.md
abiogenesis-now: good call

abiogenesis-now:
@jessicayuen Fixed--those were good points, thank you!

PR #113, opened 2017-11-20
Provide more flexibility to param schemas

### LINKED ISSUE #45: Improve overall handling of complex jsonnet param schemas
- [ ] Support top of file imports
- *Edit: Decided not to do for milestone* Consider more RHS support. (We might decide, e.g., to allow simple +, or not.)
- *Edit: Decided not to do for milestone* Support more than primitive param types.
- [ ] Support multi-line string literals.

jessicayuen:
Currently, the schema for the params.libsonnet files are very strict.
Ex: No support for top-of-file imports.

With this change, we increase the flexibility of the schema, by
recursively searching for the 'components' object node, instead of
expecting the node to be at a certain location.

Partially fixes #45

### DIFF
--- a/metadata/params/visitor.go
+++ b/metadata/params/visitor.go
@@ -1,0 +1,267 @@
+// Copyright 2017 The kubecfg authors
+//
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


metadata/params/visitor.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil, err
+	}
+	if componentsObj == nil {
+		return nil, fmt.Errorf("Could not find object node: %s", componentsID)
+	}
+
+	return componentsObj, nil
+}
+
+func visit(node ast.Node, componentsObj **ast.Object) error {
```
hausdorff: Two things:

1. I think having 2 visitor functions is super fraught, because the Jsonnet AST is pretty complex, and we do anticipate the AST evolving significantly. I really think we should try to generalize the one we have instead of having 2.
1. I would greatly prefer to not have `**ast.Object` if we can avoid it. In the first case, it means you have to do a `nil` check for both the pointer and the pointer that's being pointed to.

jessicayuen:
Going to simplify this in a follow-up PR to only handle top-of-file imports, as the current approach can be error-prone.

PR #128, opened 2017-11-21
Better support for jsonnet param schemas

### LINKED ISSUE #45: Improve overall handling of complex jsonnet param schemas

jessicayuen:
Fixes #45 

- Supports top-level imports for component params schema
- Support reading of multi-line jsonnet params

### DIFF
--- a/metadata/params/params_test.go
+++ b/metadata/params/params_test.go
@@ -296,6 +296,23 @@ func TestGetComponentParams(t *testing.T) {
 }`,
 			Params{"name": `"foo-bar"`, "replicas": "1"},
 		},
+		// Test case where one of the param values is a block string
+		{
+			"foo",
+			`
+{
+	components: {
+		"foo": {
+			name: |||
+        name
+        is
+        foo
+			|||,
+		}
+	},
+}`,
+			Params{"name": "|||\nname\nis\nfoo\n|||"},
+		},
 	}
 
 	errors := []struct {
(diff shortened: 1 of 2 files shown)


metadata/params/params_test.go
```diff
@@ -119,6 +119,41 @@ func TestAppendComponentParams(t *testing.T) {
 }`,
 			Params{"replicas": "5", "name": `"foo-bar"`},
 			`
+{
```
bryanl: I'd suggest that these are getting big enough where they should be moved into their own separate files and loaded as fixtures. The test data makes it difficult to see what test is being run.

bryanl: Also, it might be good to take advantage of go subtests as well. This will allow for more descriptive failures instead of the whole testing failing. e.g:

```go
cases := []struct{
  name string
}{}

for _, case := range cases {
  t.Run(string, func(t *testing.T) {
    // where the magic happens
  })
}
```

jessicayuen: Thanks for the feedback @bryanl  -- I'll open a ticket to do the restructuring of these tests as they can be applied to a few areas of the codebase. #168

PR #146, opened 2017-11-22
Improve logging for init

### LINKED ISSUE #115: Improve output of `ks init`
This is what I have now
 ```
$ ksonnet init config
INFO  Using context 'foo'
INFO  Creating environment 'default' with namespace '', pointing at server at address 'https://1.2.3.4'
INFO  Generating environment metadata at path '/Users/jbeda/tmp/ks/config/environments/default'
```

Somethings to improve this:
* Be clear about what is actually being done.  Something like "Creating a new app called "foo" in directory "bar" or some such
* Have a clear call to action.  Perhaps an example with `ks gen` and `ks apply`
* "using context" should also specify the kubeconfig file we are referencing.  "Using context xyz from config file abc".
* If people screw up here we should help them fix it.  Either by having them blow it away or giving them the commands.  Example: "To undo this simply delete directory xyz and rerun `ks init`.  Use flag --xyz to set the context or --abc to point at a different config file.  Try `ks init --help` for more options.

### PR DESCRIPTION
**Pull request by jessicayuen**
Fixes #115 

- Add message:  Creating a new app 'foo' at path '/path/to/foo'
- On success, add message:  ksonnet app successfully created! Next, try
  creating a component with `ks generate`
- On failure, provide suggestions for the user.
- Make note that the context is retrieved from the kubeconfig file at
  the environment variable $KUBECONFIG

### DIFF
--- a/metadata/manager.go
+++ b/metadata/manager.go
@@ -146,14 +146,14 @@ func initManager(name string, rootPath AbsPath, spec ClusterSpec, serverURI, nam
 
 	// Initialize user dir structure.
 	if err := m.createUserDirTree(); err != nil {
-		return nil, err
+		return nil, errorOnCreateFailure(name, err)
 	}
 
 	// Initialize environment, and cache specification data.
 	if serverURI != nil {
 		err := m.createEnvironment(defaultEnvName, *serverURI, *namespace, extensionsLibData, k8sLibData, specData)
 		if err != nil {
-			return nil, err
+			return nil, errorOnCreateFailure(name, err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: metadata/manager.go [hunk 1]
```diff
@@ -453,3 +455,7 @@ components + {
 }
 `)
 }
+
+func errorOnCreateFailure(appName string, err error) error {
+	return fmt.Errorf("%s\nTo undo this simply delete directory '%s' and re-run `ks init`. If error persists, try using flag '--context' to set a different context or run `ks init --help` for more options", err, appName)
```
abiogenesis-now: suggestion: put a newline after "re-run `ks init`".
grammar nit: s/If error/If the error

pkg/kubecfg/init.go
```diff
@@ -24,5 +27,8 @@ func NewInitCmd(name string, rootPath metadata.AbsPath, specFlag string, serverU
 
 func (c *InitCmd) Run() error {
 	_, err := metadata.Init(c.name, c.rootPath, c.spec, c.serverURI, c.namespace)
+	if err == nil {
+		log.Info("ksonnet app successfully created! Next, try creating a component with `ks generate`")
```
abiogenesis-now: nice! I like this. 
nit: period at the end of this.

jessicayuen: I'll add a period here. 

Note: Not sure why stylistically we haven't been adding punctuation to the end of our logging messages -- we might want to go through this later...

Issue #49, opened 2017-11-06
Check UX for namespace scenarios


abiogenesis-now:
example: What happens when an environment has namespace `test` but the `test` namespace doesn't exist on the cluster?

(and other such flows)

Make sure that error messaging is appropriate.

jessicayuen:
I've tested these scenarios over while fixing #148, so I'm going to bump off p0 but keep this ticket open for things I may have missed..

PR #148, opened 2017-11-22
Initialize namespace to 'default' if not specified

### LINKED ISSUE #116: Always set namespace with `ks init`
Right now if you run `ks init` with no explicit namespace set then no namespace will be saved into the environment. For the purposes of reproducibility we should *always* pick a namespace.  If a user doesn't specify one during `ks init` or the context is blank, then default to `default` and perhaps give the user a warning.

jessicayuen:
If the namespace is not provided in the target kubeconfig context, use
'default'.

Also squashes a bug where the namespace, server flags may have been
ignored during init.

Fixes #116, #49

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -105,33 +105,9 @@ var envAddCmd = &cobra.Command{
 
 		name := args[0]
 
-		server, namespace, context, err := commonEnvFlags(flags)
+		server, namespace, err := resolveEnvFlags(flags)
 		if err != nil {
 			return err
-		}
-
-		if len(namespace) == 0 {
-			namespace = defaultNamespace
-		}
-
-		if len(server) == 0 {
-			// If server is not provided, use the provided context.
-			// If context is also not provided, use the current context.
-			var ctx *string
-			if len(context) != 0 {
-				ctx = &context
-			}
-
-			var ns string
-			server, ns, err = resolveContext(ctx)
-			if err != nil {
-				return err
-			}
-
-			// If namespace is not provided, use the default namespace provided in the context.
-			if !flags.Changed(flagEnvNamespace) {
-				namespace = ns
-			}
 		}
 
 		appDir, err := os.Getwd()
@@ -296,22 +272,7 @@ var envSetCmd = &cobra.Command{
 			return err
 		}
 
-		server, namespace, context, err := commonEnvFlags(flags)
-		if err != nil {
-			return err
-		}
-
-		// If namespace flag is provided but without a value, use the default namespace.
-		if flags.Changed(flagEnvNamespace) && len(namespace) == 0 {
-			namespace = defaultNamespace
-		}
-
-		if len(context) != 0 {
-			server, _, err = resolveContext(&context)
-			if err != nil {
-				return err
-			}
-		}
+		server, namespace, err := resolveEnvFlags(flags)
 
 		c, err := kubecfg.NewEnvSetCmd(originalName, name, server, namespace, manager)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


hausdorff:
@jessicayuen I think this has to be rebased, right?

jessicayuen:
@hausdorff  -- yes, done.

cmd/env.go
```diff
@@ -361,3 +337,38 @@ func commonEnvFlags(flags *pflag.FlagSet) (server, namespace, context string, er
...
+			return "", "", err
+		}
+
+		if flags.Changed(flagEnvNamespace) {
+			// if a namespace is specified, we don't want to use the default one
+			namespace = ns
+		}
+	} else {
+		if len(ns) == 0 {
+			namespace = defaultNamespace
```
hausdorff: By my count we're setting `namespace = defaultNamespace` in 3 different places. It's kind of a maze right now (and it's my own fault).

I feel like we're going to shoot ourselves in the foot soon if we don't consolidate this into one place.

cmd/env.go
```diff
@@ -361,3 +337,38 @@ func commonEnvFlags(flags *pflag.FlagSet) (server, namespace, context string, er
...
+	if err != nil {
+		return "", "", err
+	}
+
+	if len(server) == 0 {
+		// If server is not provided, use the provided context.
+		// If context is also not provided, use the current context.
+		var context *string
+		if len(ctx) != 0 {
+			context = &ctx
```
hausdorff: This block of code seems odd! We're setting `context = &ctx` so that `resolveContext` can then do a similar dance and set `ctxName = *context`. I feel like I'm on the edge of being able to understand this code at all. Do you think it's possible to simplify this so that it's less ... intricate?

jessicayuen: I've cleaned this up by separating `resolveContext` into 

1. `resolveContext()` which resolves based on the default context and
2. `resolveContextFor(context string)` which resolves based on the context in the argument.

That should hopefully clear things up.

cmd/env.go
jessicayuen: @hausdorff I've cleaned up the code a little, please take another look when you get the chance.

cmd/env.go
```diff
@@ -361,3 +322,32 @@ func commonEnvFlags(flags *pflag.FlagSet) (server, namespace, context string, er
 
 	return server, namespace, context, nil
 }
+
+func resolveEnvFlags(flags *pflag.FlagSet) (server, namespace string, err error) {
```
bryanl: Named returns coupled with a naked return are hard to reason about in functions longer than a few lines. In this case, these aren't needed since you are returning concrete values below.

cmd/env.go
```diff
@@ -361,3 +322,32 @@ func commonEnvFlags(flags *pflag.FlagSet) (server, namespace, context string, er
...
+		}
+	}
+
+	if len(ns) == 0 {
+		namespace = defaultNamespace
+	} else {
+		namespace = ns
+	}
+
+	return
```
bryanl: Instead of the naked return, do a `return server, ns, nil` and change the conditional above to 

```go
if ns == "" {
  ns = defaultNamespace
}

File: cmd/root.go [hunk 5]
```diff
@@ -134,33 +134,42 @@ func namespaceFor(c clientcmd.ClientConfig, overrides *clientcmd.ConfigOverrides
 	return ns, err
 }
 
-// resolveContext returns the server and namespace of the cluster at the provided
-// context. If context is nil, the current context is used.
-func resolveContext(context *string) (server, namespace string, err error) {
+// resolveContext returns the server and namespace of the cluster at the
+// default context.
+func resolveContext() (server, namespace string, err error) {
```
bryanl: This isn't a blocking comment, but you could combine `resolveContext` and `resolveContextFor` since as a user of this internal API, I know what I want, and I may or may not have a context string. This would simplify the internal API and reduce the duplicate a tiny bit.

jessicayuen:
@bryanl Good feedback. Fixed, PTAL.

PR #153, opened 2017-11-22
💪 Make env-name a required param in "env commands"

### LINKED ISSUE #100: Remove the standalone `-f` flag
Offline discussions with @hausdorff:

The current version of `ksonnet.next` is structured heavily around the ksonnet application directory structure and the concept of environments. Should we continue to support the `show`, `apply`, `delete`, etc. on standalone files outside the application directory structure?

### LINKED ISSUE #147: `base.libsonnet` should import JSON and YAML files
Right now it ignores them.

hausdorff:
Fixes #100.

This commit will transition the ksonnet CLI away from accepting either a
list of files or an environment (or both). A detailed explanation
follows.

Historically, the ksonnet CLI has had several "environment commands".
These commands (e.g., `apply`, `diff`, `delete`, `show`, etc.) all took
one of the following:

  1. An environment name. For example, `apply us-west/dev` will `apply`
     everything in the `components/` directory to the cluster denoted by
     `us-west/dev`.
  2. A set of files. For example, `apply -f foo.jsonnet -f bar.jsonnet`
     would apply those two specific Jsonnet files to the server
     specified by the current context in $KUBECONFIG.
  3. Both an environment name and a set of files. For example,
     `apply us-west/dev -f foo.jsonnet -f bar.jsonnet` would `apply`
     those two files to the cluster denoted by the `us-west/dev`
     environment.

This "duality" remained in place primarily because it was important for
the ksonnet CLI to maintain functional compatibility (though not strict
CLI-level app compatibility) with the kubecfg tool.

Some time ago the ksonnet CLI was forked from the kubecfg CLI, but it is
only in this commit that we abandon this duality. Specifically, this
commit will _require_ the environment name for all commands. For
example, the form of `apply` will now be:
`apply <env-name> [-f <file>]`.

There are main parts to this:

  1. Refactoring core abstractions that supports this "duality".
     Specifically, removing the `cmd.envSpec` type and several
     functions, such as `cmd.parseEnvCmd`.
  2. Updating the tests to reflect this new directory structure.


cmd/root.go
```diff
@@ -454,21 +417,60 @@ func expandEnvCmdObjs(cmd *cobra.Command, envSpec *envSpec, cwd metadata.AbsPath
...
 			continue
 		}
 
 		name := strings.TrimSuffix(path.Base(p), ext)
 		name = params.SanitizeComponent(name)
-		fmt.Fprintf(&obj, "  %s: import \"%s\",\n", name, p)
+
+		// Filter! If the filter has more than 1 element and the component name is
+		// not in the filter, skip.
+		if _, exists := filter[name]; len(filter) > 0 && !exists {
```
jessicayuen: The `len(filter) > 0` check here is a little confusing to me. Can it be pulled out, before the loop?

Can we also put this specific logic at the top of the loop -- so that we can ignore all the generating import statement logic if the check passes?

Makefile
```diff
@@ -51,7 +51,7 @@ gotest:
 
 jsonnettest: ks $(JSONNET_FILES)
 #	TODO: use `ks check` once implemented
-	./ks -J lib show -f $(KCFG_TEST_FILE) -f $(GUESTBOOK_FILE) >/dev/null
+	cd testdata/test-app && ../../ks -J lib show default -c $(KCFG_TEST_FILE) -c $(GUESTBOOK_FILE) >/dev/null
```
jessicayuen: What do you think about calling one or the other `test-data` or `testapp` for consistency?

cmd/root.go
```diff
@@ -300,33 +300,10 @@ func restClientPool(cmd *cobra.Command, envName *string) (dynamic.ClientPool, di
...
-		return nil, err
-	}
-
-	var env *string
-	if len(args) == 1 {
-		env = &args[0]
-	}
-
-	return &envSpec{env: env, files: files}, nil
+	cmd.PersistentFlags().StringArrayP(flagComponent, flagComponentShort, nil, "Filename or directory that contains the configuration to apply (accepts YAML, JSON, and Jsonnet)")
```
jessicayuen: description needs updating, see below.

cmd/validate.go
```diff
@@ -36,10 +36,12 @@ var validateCmd = &cobra.Command{
 	Use:   "validate [env-name] [-f <file-or-dir>]",
 	Short: "Compare generated manifest against server OpenAPI spec",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) > 1 {
+		if len(args) != 1 {
 			return fmt.Errorf("'validate' takes at most a single argument, that is the name of the environment")
```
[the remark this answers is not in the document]
hausdorff: Fixed this in a couple other commands too.

testdata/test-app/environments/default/spec.json
```diff
@@ -0,0 +1,4 @@
+{
+  "server": "example.com",
+  "namespace": "example-ns"
+}
```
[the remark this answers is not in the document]
hausdorff: This is auto-generated. I think we'll not add a newline.

abiogenesis-now:
@jessicayuen on radar, i'm think instead of the giant PR i have out, i'm going to submit CLI doc revisions one `ks` command at a time

jessicayuen:
lgtm, pending travis fixes!

jessicayuen:
The feature changes in this PR are approved, however outstanding work will be continued in #174 due to the time sensitivity of this change, and no access to modify the branch with which these changes are committed. I've ported the commit with the correct author to the new PR. I have to close this PR for that reason to not merge duplicate changes.

PR #162, opened 2017-11-22
Use relative paths on jsonnet imports

### LINKED ISSUE #149: Generated files should not use absolute paths in `imports` (or generally)
For example, `main.jsonnet` contains an absolute path pointing to `base.libsonnet`. Not good, makes it un-portable between machines.

jessicayuen:
Currently absolute paths are being used for jsonnet imports. This makes
it difficult to share the files across different systems.

Fixes #149


jessicayuen:
Agree @hausdorff, but I think it's an enhancement that we can address after the release. Prefer to get this simple fix checked in so it doesn't block file portability issues between machines.

PR #166, opened 2017-11-27
 Clarify errors for commands requiring subcommands

### LINKED ISSUE #140: `ks pkg get` displays unhelpful error message
```
bash-3.2$ ks pkg get incubator@redis:test-reg
ERROR Command 'pkg' requires a subcommand

Usage:
  ks pkg [flags]
  ks pkg [command]

Available Commands:
  describe    Describe a ksonnet package
  install     Install a package as a dependency in the current ksonnet application
  list        Lists information about all dependencies known to the current ksonnet app

Flags:
  -h, --help   help for pkg

Global Flags:
  -v, --verbose count[=-1]   Increase verbosity. May be given multiple times.

Use "ks pkg [command] --help" for more information about a command.
```

jessicayuen:
Currently, commands such as `ks env non-existent-subcommand` will output
an error `Command 'env' requires a subcommand`.

This is not helpful since the user may assume there is the subcommand
'non-existent-subcommand'. This commit will clarify the error messages
in this scenario.

Fixes #140

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -18,6 +18,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -90,6 +91,9 @@ uniquely identifies the cluster.
 		  staging.jsonnet
           params.libsonnet`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 0 {
+			return fmt.Errorf("%s is not a valid subcommand\n\n%s", strings.Join(args, " "), cmd.UsageString())
+		}
 		return fmt.Errorf("Command 'env' requires a subcommand\n\n%s", cmd.UsageString())
 	},
 }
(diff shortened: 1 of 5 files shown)


jessicayuen:
Sorry for being unclear!

This change just differentiates between the following scenarios:

1. `ks env addition`
2. `ks env`

In the first scenario, the output will be:
```
➜  foo2 git:(pkg-get-help) ✗ ../ks env test
ERROR test is not a valid subcommand

Usage:
  ks env [flags]
  ks env [command]

Available Commands:
  add         Add a new environment to a ksonnet project
  list        List all environments in a ksonnet project
  rm          Delete an environment from a ksonnet project
  set         Set environment fields such as the name, server, and namespace.
```

vs:
```
➜  foo2 git:(pkg-get-help) ✗ ../ks env     
ERROR Command 'env' requires a subcommand

Usage:
  ks env [flags]
  ks env [command]

Available Commands:
  add         Add a new environment to a ksonnet project
  list        List all environments in a ksonnet project
  rm          Delete an environment from a ksonnet project
  set         Set environment fields such as the name, server, and namespace.
```

Note the difference in the first lines of the error (before the usage). It's subtle but currently on master both output the second scenario.

PR #167, opened 2017-11-28
pkg install: Use 'master' as refSpec where version isn't provided

### LINKED ISSUE #151: Remove dependency on `test-reg` as version for registry

jessicayuen:
Fixes #151 

Currently, if a version is not provided to the following `pkg install`
command

`../ks pkg install incubator/redis`

, the following error will occur:

`ERROR GET
https://api.github.com/repos/ksonnet/parts/commits/incubator/redis: 404
Not Found []`

This commit will pull in the latest package from master where the
version is not specified.

Also set the default repository ref to 'master' instead of 'test-reg'.

### DIFF
--- a/cmd/pkg.go
+++ b/cmd/pkg.go
@@ -249,7 +249,10 @@ func parseDepSpec(cmd *cobra.Command, spec string) (registry, libID, name, versi
 	if len(split) > 2 {
 		return "", "", "", "", fmt.Errorf("Symbol '@' is only allowed once, at the end of the argument of the form <registry>/<library>@<version>")
 	}
-	version = split[len(split)-1]
+	version = ""
+	if len(split) == 2 {
+		version = split[1]
+	}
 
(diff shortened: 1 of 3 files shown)


cmd/pkg.go
```diff
@@ -249,7 +249,10 @@ func parseDepSpec(cmd *cobra.Command, spec string) (registry, libID, name, versi
 	if len(split) > 2 {
 		return "", "", "", "", fmt.Errorf("Symbol '@' is only allowed once, at the end of the argument of the form <registry>/<library>@<version>")
 	}
-	version = split[len(split)-1]
+	version = ""
```
bryanl: If there is no `@`, couldn't you append a `@master`? If you made that change here, you wouldn't have to change anything else.

jessicayuen: Nice catch -- it's fixed! Please take another look.

PR #175, opened 2017-11-28
[docs] Remove diff command in example

abiogenesis-now:
Since `ks diff` output is not very clean or user-friendly, let's show `ks param list` instead.

@jessicayuen PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,8 @@ Now let's try upgrading the container image to a new version:
 # Bump the container image to a different version
 ks param set guestbook-ui image alpinejay/dns-single-redis-guestbook:0.4
 
-# See the differences between your local manifest and what's running on your cluster
-# (ERROR is expected here since there are differences)
-ks diff local:default remote:default
+# View updated param values
+ks param list
 
 # Update your cluster with your latest changes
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,9 +101,8 @@ Now let's try upgrading the container image to a new version:
 # Bump the container image to a different version
 ks param set guestbook-ui image alpinejay/dns-single-redis-guestbook:0.4
 
-# See the differences between your local manifest and what's running on your cluster
-# (ERROR is expected here since there are differences)
-ks diff local:default remote:default
+# View updated param values
+ks param list
```
jessicayuen: What's your opinion on removing this command altogether? I think this may be of little use to the user and interrupt the flow, since the next step after `apply` is for them to visually see the changes.

Minor thing though. Approving - leaving this to your discretion; I'd be okay with either option.

abiogenesis-now: Opting for keeping it in---the formatting for this command is actually quite nice, and it kind of reinforces the fact (implicitly) that state changes like these are tracked locally. (I do see what you mean though! If this trips anyone else up, I'll be sure to change).

PR #176, opened 2017-11-29
Add support for custom directory location for ks app

### LINKED ISSUE #114: Rethink `ks init` parameters
`ks init` takes an argument.  In many cases this will be the subdir of the application itself.  

Example: I have a github org and I want the ks dir to be under that. In that case, "application name" isn't quite right.  "config dir name" is probably more appropriate.  But it is more than just the dir name as this name gets put into the "name" field in app.yaml.

Specifically: if I have an app called "caribou" I'll probably call `ksonnet init` from the root of the github repo for caribou.  If I call `ks init caribou` I'll have a directory called "caribou/caribou" most probably.  That isn't great. But if I call `ks init k8s-config` then that sets the app name in app.yaml and we may key off of that for automatic labels and such at some point.

Perhaps we should have two arguments to `ks init`.  One that is the app name.  The second is the directory to create. We can pick good defaults for both. App name is the basename of the current directory.  The directory is something like ksconfig or something else sane.

bryanl:
* If custom dir is relative, it is created relative to CWD
* if custom dir is absolute, it is created at the specificed location
* if custom dir is not specified, a directory created at CWD/app-name is created

Fixes #114


cmd/init.go
```diff
@@ -49,7 +55,18 @@ var initCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		appRoot := metadata.AbsPath(path.Join(appDir, appName))
+
+		initDir, err := flags.GetString("dir")
```
jessicayuen: refer to the constant `flagInitDir` here instead of `"dir"`

cmd/init.go
```diff
@@ -49,7 +55,18 @@ var initCmd = &cobra.Command{
...
+		if err != nil {
+			return err
+		}
+
+		path, err := genKsRoot(appName, appDir, initDir)
```
jessicayuen: What's your opinion on renaming `appDir` to `wd` for working directory? (since it's potentially no longer the app dir) and calling `initDir` the `appDir`?

cmd/init.go
```diff
@@ -122,3 +139,24 @@ ks init app-name --api-spec=version:v1.7.1
 # specific build of Kubernetes to generate 'ksonnet-lib'.
 ks init app-name --api-spec=file:swagger.json`,
 }
+
+func genKsRoot(appName, appDir, ksDir string) (string, error) {
+	if appDir == "" {
+		return "", errors.New("invalid application directory")
```
jessicayuen: related to above, this might be `"invalid working directory"`. Consider renaming `appDir` and `ksDir` to `wd` and `appDir` respectively.

cmd/init.go
```diff
@@ -122,3 +139,24 @@ ks init app-name --api-spec=version:v1.7.1
...
+	if appName == "" && ksDir == "" {
+		return "", errors.New("invalid application name")
+	}
+
+	if ksDir != "" {
+		if filepath.IsAbs(ksDir) {
+			return ksDir, nil
+		}
+
+		path := fmt.Sprintf("%s%s%s", appDir, string(filepath.Separator), ksDir)
```
jessicayuen: Probably better to use `return filepath.Join(appDir, ksDir)` here in order to not deal with trailing slashes

PR #219, opened 2017-12-12
Accurately read/write non-ASCII param identifiers

### LINKED ISSUE #214: `param set` incorrectly supporting hyphenated param names
Ex:
`ks param set guestbook-ui foo-bar 5`

is outputting jsonnet
```
  components: {
    // Component-level parameters, defined initially from 'ks prototype use ...'
    // Each object below should correspond to a component in the components/ directory
    "guestbook-ui": {
      foo-bar: 20,
    },
  },
```

`foo-bar` needs to be quoted.

jessicayuen:
**Fixes #214**

Currently, if we run `param set guestbook-ui foo-bar x`,

The jsonnet snippet will write: `foo-bar: "x"`.

However, this is incorrect jsonnet syntax. This commit will fix this
issue, to instead write `"foo-bar": "x"`.

Also updates the `Get...` Param functions to return non-quoted
identifiers for prettier display.

### DIFF
--- a/utils/strings.go
+++ b/utils/strings.go
@@ -17,6 +17,7 @@ package utils
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 
 	"github.com/PuerkitoBio/purell"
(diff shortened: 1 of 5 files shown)


utils/strings.go
```diff
@@ -34,6 +35,14 @@ func IsASCIIIdentifier(s string) bool {
 	return true
 }
 
+// QuoteNonASCII puts quotes around a identifiers that contain non-ASCII characters.
```
jessicayuen: note to self for wednesday: grammar!

PR #220, opened 2017-12-12
Check error while enumerating environments

tanner-bruce:
It is possible for the walk function to be called with err set


hausdorff:
Ok, so `err` here binds to the parameter passed into the anonymous function, right? Good catch. I guess `errcheck` doesn't find stuff like this, which is unfortunate.

tanner-bruce:
Yup, it looks like if `afero.Walk` fails while listing the dir, err will be set. I found it via `staticcheck`. `gometalinter` also provides a nice laundry list! :)

PR #228, opened 2017-12-15
Implement command `ks registry add`

jessicayuen:
Fixes #226

Currently users are unable to add their own registries through the CLI.
This limits them to a small subset of prototypes found in the default
incubator registry.

This commit will add the command `ks registry add`, that allows users to
add registries supporting the `github` protocol.

It will be of the form `ks registry add <registry-name> <registry-uri>
[--version]`. If a version is not specified, `latest` will be used.

PR #230, opened 2017-12-15
Allow the default registry to be configured via environment variable.

jlewi:
* This unblocks projects (like [Kubeflow](google/kubeflow)) that want to start using their
  own registry.


jlewi:
Looks like #228 is already pending to add ks registry add which is even better.

PR #245, opened 2017-12-18
Test all branches of GH URI-parsing code

hausdorff:
When a user adds a registry (e.g., through a command like `ks registry
add`, or implicitly through `ks init`) we must parse a URI pointing at
registry hosted on github.com, since we currently only support the
`"github"` protocol.

This logic is somewhat complex, since we must

  1. infer the location of a `registry.yaml` file given the URI, and
  2. accept and parse a broad set of URIs a user might provide.

For example, consider the following valid URIs, and what we do to infer
the location of the `registry.yaml` file that specifies a registry:

  URIs with an explicit `registry.yaml`:
    github.com/exampleOrg/exampleRepo/blob/master/registry.yaml
    github.com/exampleOrg/exampleRepo/blob/master/incubator/registry.yaml

  URIs with an implicit `registry.yaml`:
    github.com/exampleOrg/exampleRepo/
    github.com/exampleOrg/exampleRepo/tree/master

  URIs with different protocols (or no protocol specified)
    github.com/exampleOrg/exampleRepo
    http://github.com/exampleOrg/exampleRepo
    https://github.com/exampleOrg/exampleRepo
    www.github.com/exampleOrg/exampleRepo

And so on.

Because this parsing logic has many branches, it is important for us to
test every branch, so that the user is not surprised when a command like
`registry add` doesn't work.

This commit will introduce such tests.

### DIFF
--- a/metadata/registry_test.go
+++ b/metadata/registry_test.go
@@ -2,11 +2,163 @@ package metadata
 
 import (
 	"path"
+	"testing"
 
 	"github.com/ksonnet/ksonnet/metadata/app"
 	"github.com/ksonnet/ksonnet/metadata/parts"
 	"github.com/ksonnet/ksonnet/metadata/registry"
 )
+
+func TestParseGiHubRegistryURITest(t *testing.T) {
+	tests := []struct {
+		// Specification to parse.
+		uri string
+
+		// Optional error to check.
+		targetErr error
+
+		// Optional results to verify.
+		targetOrg                  string
+		targetRepo                 string
+		targetRefSpec              string
+		targetRegistryRepoPath     string
+		targetRegistrySpecRepoPath string
+	}{
+		//
+		// `parseGitHubURI` should correctly parse org, repo, and refspec. Does not
+		// test path parsing.
+		//
+		{
+			uri: "github.com/exampleOrg1/exampleRepo1",
+
+			targetOrg:                  "exampleOrg1",
+			targetRepo:                 "exampleRepo1",
+			targetRefSpec:              "master",
+			targetRegistryRepoPath:     "",
+			targetRegistrySpecRepoPath: "registry.yaml",
+		},
+		{
+			uri: "github.com/exampleOrg2/exampleRepo2/tree/master",
+
+			targetOrg:                  "exampleOrg2",
+			targetRepo:                 "exampleRepo2",
+			targetRefSpec:              "master",
+			targetRegistryRepoPath:     "",
+			targetRegistrySpecRepoPath: "registry.yaml",
+		},
+		{
+			uri: "github.com/exampleOrg3/exampleRepo3/tree/exampleBranch1",
+
+			targetOrg:                  "exampleOrg3",
+			targetRepo:                 "exampleRepo3",
+			targetRefSpec:              "exampleBranch1",
+			targetRegistryRepoPath:     "",
+			targetRegistrySpecRepoPath: "registry.yaml",
+		},
+		{
+			// Fails because `blob` refers to a file, but this refers to a directory.
+			uri:       "github.com/exampleOrg4/exampleRepo4/blob/master",
+			targetErr: errInvalidURI,
+		},
+		{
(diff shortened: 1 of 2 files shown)


hausdorff:
This commit thoroughly tests `metadata.parseGitHubURI`. A few words about why this PR is important, and why I think it is a big step towards fully-specified behavior in tests.

The crux of the `registry add` and and library resolution logic is specified by the interface `registry.Manager`:

```
type Manager interface {
	RegistrySpecDir() string
	RegistrySpecFilePath() string
	FetchRegistrySpec() (*Spec, error)
	MakeRegistryRefSpec() *app.RegistryRefSpec
	ResolveLibrarySpec(libID, libRefSpec string) (*parts.Spec, error)
	ResolveLibrary(libID, libAlias, version string, onFile ResolveFile, onDir ResolveDirectory) (*parts.Spec, *app.LibraryRefSpec, error)
}
```

Of these, `Manager#RegistrySpecDir`, `Manager#RegistrySpecFilePath` and `Manager#MakeRegistryRefSpec` are completely trivial, simply returning values generated after calling `metadata.parseGitHubURI`.

Each of the remaining methods likewise depends critically on `metadata.parseGitHubURI`. Specifically:

* `Manager#FetchRegistrySpec` is a thin layer around the GitHub client call that retrieves `registry.yaml` from the GitHub repository, and deserializes it. The challenge of this work is actually done by `metadata.parseGitHubURI`, which supplies the information the GitHub client needs to uniquely identify a specific version of this file (including inferring the location in the repository).
* `Manager#ResolveLibrarySpec` is a thin layer around the GitHub client call that retrieves the `parts.yaml` file from the GitHub repository, and deserializes it. Again the challenging work is actually done by `metadata.parseGitHubURI`, which supplies the information the GitHub client needs to uniquely identify a specific version of this file (including inferring the location in the repository).
* `Manager#ResolveLibrary` is less trivial. It recursively retrieves the entire directory of a library, repeatedly using the GitHub client to retrieve directories and files. This still depends on `metadata.parseGitHubURI` to supply the information required to actually recursively explore the directory.

Of these, `Manager#ResolveLibrary` is most likely to be wrong. The problem with including tests against the GitHub API directly is that GitHub imposes a rate limit, which makes automated testing quite challenging. Eventually when we support local repositories, we may be able to clone and store the repository in some scratch space, and then include a simple repository for testing, which would get us a lot closer to total test coverage here.

metadata/registry_test.go
```diff
@@ -2,12 +2,163 @@ package metadata
...
+						t.Fatalf("Expected URI '%s' parse to fail with err '%v', got: '%v'", uri, test.targetErr, err)
+					}
+					continue
+				}
+
+				if err != nil {
+					t.Fatalf("Expected URI '%s' parse to succeed, but failed iwth error '%v'", uri, err)
+				}
+
+				if org != test.targetOrg {
```
bryanl: if this or any of the tests below it fail, you won't immediately know what URI the error corresponds to.  If you wrapped it in a `t.Run(uri, func(t *testing.T){...})`, you'd have a better clue.

metadata/registry_test.go
```diff
@@ -2,12 +2,163 @@ package metadata
...
+	for _, test := range tests {
+		// Make sure we correctly parse each URN as a bare-domain URI, as well as
+		// with 'http://' and 'https://' as prefixes.
+		for _, prefix := range []string{"http://", "https://", "http://www.", "https://www.", "www.", ""} {
+			// Make sure we correctly parse each URI even if it has the optional
+			// trailing `/` character.
+			for _, suffix := range []string{"/", ""} {
+				uri := prefix + test.uri + suffix
+
+				org, repo, refspec, registryRepoPath, registrySpecRepoPath, err := parseGitHubURI(uri)
```
bryanl: This function returns five values and an error. Are these values their own type?

hausdorff: Nope.

metadata/registry_test.go
hausdorff: Ok.

Issue #197, opened 2017-12-03
Consider switching to dep from govendor


jbeda:
It is the new hotness and seems to be what folks are converging on...

kstewart:
... and also what we have internally decided to standardize on at Heptio.

jessicayuen:
May want to hold off on this one until #215 is addressed.

anguslees:
See also open issues like https://github.com/kubernetes/client-go/issues/318

Last time I tried (a few weeks ago), `dep` cost me a day of fiddling and then I went back to `govendor`.  I would be delighted to hear that things have improved.

PR #265, opened 2018-01-04
convert from govendor to dep

bryanl:
Convert from govendor to dep. 

_note_: this updates a few packages as well.

* [x] convert to dep
* [x] updating contrib docs


hausdorff:
I think it's worth moving to `dep`, but it's hard for me to tell how confident we are in these changes.

1. Does this PR complete the transition to `dep`? (I don't see a message like "Fixes #197".) If not, what remains?
1. Did we resolve the problems Gus had when he tried to transition over ksonnet/kubecfg (as described in #197)?

bryanl:
@hausdorff those issues haven't presented themselves in this change. Also, we can potentially update `client-go` to the latest release (6.x) and that should solve the dep/client-go issues raised in #197.

bryanl:
@hausdorff Also, yes this completes the transition to `dep`.

hausdorff:
@bryanl We should upgrade `client-go`, but I think we should do it in a different review. It's a pretty big operational risk (1) because we have accumulated a lot of _ad hoc_ fixes for bugs in the current version, and (2) they do break the internal APIs all the time.

I'll take a look at the review in a bit.

hausdorff:
@bryanl Oh, also, if we're going to transition to `dep` and upgrade `client-go`, we should amortize that risk by making sure we upgrade `kubecfg` on both those things at the same time. Bitnami has to use that code in anger, and if there are issues, they will surface quite a bit faster when vetted through there.

bryanl:
The goal was to make the changes in steps. 1) move to dep 2) update client-go and friends. This PR Is only the first step.

hausdorff:
@bryanl ok, so, just to be clear: the plan _is_ to update kubecfg as well, yes?

hausdorff:
Summarizing offline conversations: yes, we're planning on updating kubecfg.

bryanl:
@hausdorff I'm going to review to see if that is possible. dep and govendor. I may have to go back and adjust each dep by hand. The problem is explicit vs implicit dependencies.

hausdorff:
@bryanl Ok, if it's not easy, then I recommend just shipping it as-is. Another way to ameliorate this risk is to ship a release candidate for 0.9, so that at least people have a chance to test it in the wild. (I bring this up because we have a bunch of version-specific fixes for bugs in our dependencies, strung throughout the code, so. Ick.)

jessicayuen:
Just incase you miss this, should do a rebase before merging. Some commits are dirty.

jessicayuen:
I've noticed this is taking a lot longer to build (even on subsequent builds). Are you seeing similar behavior @bryanl ?

```
➜  ksonnet git:(pr-265) ✗ time make
go build -o ks -ldflags="-X main.version=foo "  .
./docs/generate/update-generated-docs.sh
make  42.67s user 4.83s system 209% cpu 22.687 total
```

bryanl:
I didn't notice. Will investigate in the morning.

Gopkg.lock
```diff
@@ -0,0 +1,313 @@
...
+  revision = "21300e3e11c918b8e6a70fb7293b310683d6c046"
+  version = "v3.0.0"
```
redbaron: this  is for 1.6 Kubernetes, does it mean that more recent features like CRDs are not supported?

PR #267, opened 2018-01-05
design proposal: ksonnet lib simple constructors

bryanl:
Introduce simple constructors for ksonnet-lib, so it's easier to build
abstractions based on ksonnet-lib.

### DIFF
--- a/design/proposals/ksonnet-lib-simple-constructors.md
+++ b/design/proposals/ksonnet-lib-simple-constructors.md
@@ -48,12 +48,18 @@ Many of the existing constructors adhere to this format. This change would make 
 
 To deprecate the existing the existing custom constructors a new constructor will be introduced, `init()`. Currently, to construct a v1beta2 Deployment, you have to do the following:
 
-```jsonnet
-local params = ...;
+```js
+local params = {
+  name: "appName",
+  replicas: 3,
+  containerPort: 80,
+  image: "nginx:latest",
+  labels: {app: "customName"},
+};
+
 local deployment = k.apps.v1beta2.deployment;
 local container = k.apps.v1beta2.deployment.mixin.spec.template.spec.containersType;
 local containerPort = container.portsType;
-local labels = {app: params.name};
 local targetPort = params.containerPort;
 
 local appDeployment = deployment
@@ -63,7 +69,7 @@ local appDeployment = deployment
     container
       .new(params.name, params.image)
       .withPorts(containerPort.new(targetPort)),
-    labels
+    params.labels
 );
 ```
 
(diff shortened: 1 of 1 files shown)


design/proposals/ksonnet-lib-simple-constructors.md
```diff
@@ -0,0 +1,109 @@
...
+  .withName(params.name)
+  .withImage(params.image)
+  .withPorts(port)
+
+local appDeployment = deployment
+  .init()
+  .mixin.metadata.withName(params.name)
+  .mixin.spec.withReplicas(params.replicas)
+  .mixin.spec.template.spec.withContainers(containers)
+  .mixin.spec.template.metadata.withLabels(labels);
```
abiogenesis-now: Maybe this an issue that can be solved with documentation for `ksonnet-lib` (the autogenerated stuff we have planned), but I just wanted to point out that the prefix of `mixin.spec.template.metadata` can be confusing to new users.

If we keep this syntax, we need to point out (e.g. via docs) what a `mixin` is (~ inheritance), and that `spec.template.metadata` are nested fields in the official Deployment spec.

Even if this user knows this, it seems to implicitly require that they are familiar with the K8s spec? Again, perhaps an HTML page with all this info ( + ctrl-F) can be useful, but we should make sure that choice is explicit.

design/proposals/ksonnet-lib-simple-constructors.md
```diff
@@ -0,0 +1,109 @@
...
+    local apiVersion = {apiVersion: "group/v1"},
+    kind:: {
+      local kind = {kind: "Kind"},
+      init() :: apiVersion + kind,
+    },
+  },
+}
+```
+
+Many of the existing constructors adhere to this format. This change would make sure all resources have a constructor that only includes the apiVersion and kind.
```
abiogenesis-now: nit: if it's not too much trouble, backticks around `apiVersion` and `kind` would be helpful

design/proposals/ksonnet-lib-simple-constructors.md
```diff
@@ -0,0 +1,109 @@
...
+  .init()
+  .mixin.metadata.withName(params.name)
+  .mixin.spec.withReplicas(params.replicas)
+  .mixin.spec.template.spec.withContainers(containers)
+  .mixin.spec.template.metadata.withLabels(labels);
+```
+
+### Backwards compatibility
+
+Existing constructors will not be removed. The inclusion of `init()` constructors. Will be the only change. This change will not impact any of the current usages of ksonnet-lib. After a period of time (at most two releases of ksonnet), the existing constructors will be removed.
```
abiogenesis-now: Rewording for clarity:

This change will not immediately impact any of the current usages of ksonnet-lib. To begin, the inclusion of `init()` constructors will be the only change. Only after a period of time (at most two releases of ksonnet) will existing constructors be removed.

abiogenesis-now: if we end up doing `ks lint` we can probably add a warning when we actually deprecate the constructors and the user's ksonnet still has them

design/proposals/ksonnet-lib-simple-constructors.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Alternatives considered
+
+Instead of introducing new constructors, missing constructors can be be added to ksonnet-lib. Also, resources with incomplete, missing, or extra required items could fixed as well.
```
abiogenesis-now: Maybe add a brief sentence why these alternatives were not chosen?

design/proposals/ksonnet-lib-simple-constructors.md
bryanl: This is true. This proposal is trying to make `ksonnet-lib` a ksonnet version of `swagger.json`. All embellishments should have in our yet to be created standard library. It's difficult to use because it allows full access to build the more useful abstractions upstream.

abiogenesis-now: Ah ok, that makes sense!

design/proposals/ksonnet-lib-simple-constructors.md
```diff
@@ -0,0 +1,109 @@
...
+  .init()
+  .withContainerPort(targetPort)
+
+local container = containersType
+  .init()
+  .withName(params.name)
+  .withImage(params.image)
+  .withPorts(port)
+
+local appDeployment = deployment
```
hausdorff: Hmm, I'm not sure this code actually works? Shouldn't it be something like (I haven't tested this code):

```
deployment
  .init() +
  deployment.mixin.metadata.withName(params.name) +
  deployment.mixin.spec.withReplicas(params.replicas) +
  deployment.mixin.spec.template.spec
    .withContainers(containers)
    .withLabels(labels);
```

You basically have to use a `+` for each "namespace". So everything in `spec.template.spec` can build builder-style constructed altogether, and the `+`'d to the builder-style constructed `spec.template`, `spec`, `metadata`, and so on. I think this gets more obvious when you set more than one property in each, like:

```
  deployment.mixin.metadata
    .withName(params.name)
    .withNamespace(params.namespace) +
  deployment.mixin.spec.template.spec
    .withContainers(containers)
    .withLabels(labels);
```

bryanl: Good catch, I'll update the proposal.

hausdorff:
@bryanl any thoughts on the points I brought up above?

bryanl:
@hausdorff I'd like to ultimately remove all the code for the custom constructors. I want to put this complexity elsewhere since there is no way to determine what a proper constructor should contain by only using the swagger definition.

hausdorff:
@bryanl You can still do that, starting by making all the constructor parameters optional, right? Then we don't have to have a release with two constructors.

bryanl:
@hausdorff The reason I suggested an new constructor is so existing code will continue to work. If we make all parameters optional, wouldn't that require changes to existing code?

hausdorff:
@bryanl No, no code changes are required:

```
$ jsonnet -e "{ foo(a):: a, bar: self.foo(3) }"
{
   "bar": 3
}
$ jsonnet -e "{ foo(a=3):: a, bar: self.foo(3) }"
{
   "bar": 3
}
$ jsonnet -e "{ foo(a=3):: a, bar: self.foo() }"
{
   "bar": 3
}
```

Do you have a counter example?

adamdecaf:
cc https://github.com/ksonnet/ksonnet-lib/issues/91

Issue #222, opened 2017-12-13
How can components inherit the namespace from the environment?


jlewi:
Since namespace is a required namespace parameter, is it possible to reference that value in my components to set the namespace?

In the prototype examples it looks like namespace is hardcoded e.g. [nginx example](https://github.com/ksonnet/parts/blob/9d78d6bb445d530d5b927656d2293d4f12654608/incubator/nginx/examples/nginx.jsonnet)

jessicayuen:
+ @bryanl Pretty important, let's put this at a p1.

abiogenesis-now:
realized private-convo slack is not super helpful, copy-pasted the text (although it's basically just a rehash of the issue title):
> also I know prototypes can specify defaults that are constants. right now redis prototype has a mandatory `namespace` field. Is it reasonable to want `namespace` to be optional, and autopopulated by the `env`’s namespace at runtime (`ks apply`) if not specified?

jessicayuen:
@jlewi  Using the same `nginx` context, the current way of working around this is to specify the component file as follows:

```
local params = std.extVar("__ksonnet/params").components.nginx;

k.core.v1.list.new([
  nginx.parts.deployment.simple(params.namespace, ...),
  nginx.parts.service(params.namespace, ...),
])
```

And subsequently in your environment's `environments/<name>/params.libsonnet` file, you can set the namespace value.

Ex, in your "dev" environment you might set `environments/dev/params.libsonnet`  to:
```
...
nginx +: {
  "namespace": "dev",
}
...
```
and in your "prod" environment you might set `environments/prod/params.libsonnet`  to:
```
...
nginx +: {
  "namespace": "dev",
}
...
```

I agree that it's not very polished, though. We should be able to reference the environment's namespace without explicitly setting it in the `params.libsonnet` file again.

PR #279, opened 2018-01-08
Design proposal: explicit environment metadata

jessicayuen:
Proposal for explicit environment metadata; allowing environment metadata (such as namespace, server URI) to be reference-able at the component and prototype manifest level.
 
Attempts to address #222

### DIFF
--- a/design/proposals/explicit-environment-metadata.md
+++ b/design/proposals/explicit-environment-metadata.md
@@ -1,0 +1,147 @@
+# Explicit Environment Metadata
+
+Status: Pending
+
+Version: Alpha
+
+## Abstract
+
+We want to introduce functionality to enable the use of environment metadata (ex: cluster namespace, Server URI) in prototype and component manifest files.
+
+## Motivation
+
+* Environments have configured metadata, such as the cluster’s namespace or server URI. Users [have requested](https://github.com/ksonnet/ksonnet/issues/222) that these fields be referenceable in both prototypes and component files.
+
+Example:
+
+Consider the following [nginx prototype](https://github.com/ksonnet/parts/blob/9d78d6bb445d530d5b927656d2293d4f12654608/incubator/nginx/examples/nginx.jsonnet) where the namespace is currently hardcoded:
+
+```jsonnet
+local k = import 'k.libsonnet';
+local nginx = import '../nginx.libsonnet';
+
+local namespace = "dev-alex";
+local appName = "nginx-app";
+
+k.core.v1.list.new([
+  nginx.parts.deployment.withServerBlock(namespace, appName),
+  nginx.parts.service(namespace, appName),
+  nginx.parts.serverBlockConfigMap(namespace, appName),
+])
+```
+ 
+This example does not lend well to the ksonnet desired model where components are environment agnostic; meaning a component should not need information on which environment the resulting Kubernetes Objects will be deployed to. Yet, it is unavoidable that these component / prototype configurations may need to reference a namespace, server URI, or other metadata.
+
+## Goals
+
+* Encourage behavior for component and prototype manifests to be environment-agnostic. 
+  Example: It is encouraged to write `nginx.parts.service(env.namespace, appName)`, but not `local namespace = "foo"` or `if (env.namespace == "foo") ... else ...`.
+* Component and prototype manifests should be able to refer to environment metadata fields, such as the environment namespace in the form of `env.namespace`.
+* Environment fields should be evaluated at run-time depending on which environment the user is trying to deploy to. 
+  Example: If the user had a `prod` environment configured to the namespace `prod-ns`, running `ks apply prod` should evaluate `env.namespace` to the value `prod-ns`.
+* Produce a second version of existing prototypes without hard-coded namespace values.
+
+## Non Goals
+
+* Access to a specific environment's metadata. 
+  Example: There should not be access to an attribute like `env.prod.namespace`. The evaluation of `env.namespace` should happen on `apply`, `diff`, or other ks functionalities that operate on an environment.
+* This proposal does not suggest to reveal the environment metadata as a part of the `ks param` operations. Environment metadata will remain discrete from component parameters. Meaning, users are currently able to discover differences in component parameters through the use of the ksonnet CLI. Running `ks param diff <env-one> <env-two>` will output differences in component parameters, such as replica count, between the two environments, however will not output differences in environment metadata such as the namespace.
+
+## Proposal
+
+### Fundamentals
+
+The change will behave similarly to how parameters operate in component and prototype manifests.
+
+Components created on the `generate` command will introduce the following line to the top of the component manifest [1]:
+
+`local env = std.extVar("__ksonnet/env");`
+
+Namespace, Server URI, and other environment metadata attributes can then be accessed using `env.namespace` and `env.server`, respectively.
+
+For example, the components generated from the nginx prototype in the Motivation section above may appear as:
+
+```jsonnet
+local env = std.extVar("__ksonnet/env");
+
+local k = import 'k.libsonnet';
+local nginx = import '../nginx.libsonnet';
+
+local appName = "nginx-app";
+
+k.core.v1.list.new([
+  nginx.parts.deployment.withServerBlock(env.namespace, appName),
+  nginx.parts.service(env.namespace, appName),
+  nginx.parts.serverBlockConfigMap(env.namespace, appName),
+])
+```
+
+The corresponding prototype manifest would appear as:
+
+```jsonnet
+local k = import 'k.libsonnet';
+local nginx = import '../nginx.libsonnet';
+
+local appName = "nginx-app";
+
+k.core.v1.list.new([
+  nginx.parts.deployment.withServerBlock(import 'env://namespace', appName),
+  nginx.parts.service(import 'env://namespace', appName),
+  nginx.parts.serverBlockConfigMap(import 'env://namespace', appName),
+])
+```
+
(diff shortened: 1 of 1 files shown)


bryanl:
Would it be possible for find a more consumable name than `ambient`?

hausdorff:
Following up from a slack conversation, @jessicayuen and I have agreed:

* This is actually not a proposal for _ambient_ namespaces, but for _explicit app namespaces_. (We already have ambient namespaces, they're provided by Kubernetes to all API objects that don't override namespaces.)

The proposal will be changed accordingly.

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,52 @@
+# Ambient Namespaces
+
+Status: Pending
+
+Version: Alpha
+
+## Abstract
+
+We want to introduce functionality to enable the use of environment metadata (ex: cluster namespace, Server URI) in prototype and component manifest files.
```
hausdorff: The idea that we would expose cluster name or server URI is somewhat controversial. (I am personally for it, but I think everyone agrees, as you mention below, having people write stuff like `if (env.name) then [...]` is considered an anti-pattern, which is why other people on the team feel strongly we should not do this.)

So: did you mean to expose all of this metadata, or did you only want the namespace? Either one is ok with me, but I think we should beef up the motivation for the other two if we're going to go that route.

FWIW, personally I think there is a good case to be made that namespace is actually special, since it is actually part of the app code (_i.e._, we specify it in the YAML), while name and server URI are not (_i.e._, they do not show up in the YAML).

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Motivation
+
+* Environments have configured metadata, such as the cluster’s namespace or server URI. Users [have requested](https://github.com/ksonnet/ksonnet/issues/222) that these fields be referenceable in both prototypes and component files.
+* Component and prototype manifests should not reference a specific environment's details. See Goals and Non Goals for more clarity on what "specific" implies.
```
hausdorff: I think the first of these is pretty good motivation, and you substantiate it well. The second is really more of a constraint and you might just defer that to the goals/non-goals.

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,52 @@
...
+  nginx.parts.serverBlockConfigMap(namespace, appName),
+])
+```
+ 
+This example does not lend well to the ksonnet desired model where components are environment agnostic; meaning a component should not need information on which environment the resulting Kubernetes Objects will be deployed to. Yet, it is unavoidable that these component / prototype configurations may need to reference a namespace, server URI, or other metadata.
+
+## Goals
+
+* Encourage behavior for component and prototype manifests to be environment-agnostic. 
+  Example: It is encouraged to write `nginx.parts.service(env.namespace, appName)`, but not `local namespace = "foo"` or `if (env.namespace == "foo") ... else ...`.
```
hausdorff: Interesting, why not something like `params.global.namespace` instead of `env.namespace`? Then you could put in the params file something like:

```
{
  global:: {
    namespace:: std.extVar("__ksonnet/global/namespace"),
  },
}
```

Or something.

This re-uses a bunch of constructs we already have, which is IMHO nicer.

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,52 @@
...
+* Component and prototype manifests should be able to refer to environment metadata fields, such as the environment namespace in the form of `env.namespace`.
+* Environment fields should be evaluated at run-time depending on which environment the user is trying to deploy to. 
+  Example: If the user had a `prod` environment configured to the namespace `prod-ns`, running `ks apply prod` should evaluate `env.namespace` to the value `prod-ns`.
```
hausdorff: Short of not including `env.name` and `env.server` I'm not sure how we can prevent people from writing conditionals on them. I suspect we're ok exposing app `namespace`, since that's a part of app code anyway.

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Non Goals
+
+The following are related but not direct goals of this specific proposal
+
+* Rewriting existing prototypes to remove hard-coded namespace values.
```
hausdorff: Ha, I think this should be an explicit goal, FWIW. That they are hard-coded is the reason we _aren't_ using ambient namespaces to begin with, right? (Since ambient namespaces are given to objects that _don't_ have the `namespace` set.)

design/proposals/ambient-namespaces.md
jessicayuen: This is an interesting idea, and would work too. My only reservations (if that) would be:

`params.global.namespace` is a lot more wordy than `env.namespace` and does it articulate well enough that we are operating on an environment's attribute? That said, how opinionated do we want a ksonnet environment to be Ex: Since namespace is a metadata field of environment, should access / customization of a namespace be hidden by it's environment? If we begin to introduce them as global params, we seem to be introducing two areas where they can be configured. Not that this is necessarily bad, but I can see it causing confusion.

The plus side of this approach would allow for params to have fields such as:

```
"guestbook-ui": {
  namespace: global.namespace
}
```

design/proposals/ambient-namespaces.md
jessicayuen: The intention was to expose all this metadata, but that was with the assumption that future `ksonnet-lib` APIs may take a server address. i.e. That it may be part of the future app code.

If this is a valid assumption, I'll beef up the motivation section with this point.

design/proposals/ambient-namespaces.md
jessicayuen: We probably have to supply the rewrites as different versions, otherwise we'll run into backwards-compatibility problems when a 0.8.0 ks user attempts to use these updates.

design/proposals/ambient-namespaces.md
hausdorff: Oh, interesting, I actually can't think of a time we'd want to have server URI exposed as part of ksonnet-lib. What sort of scenarios were you thinking of?

design/proposals/ambient-namespaces.md
hausdorff: Those are good points. I think you're right that we should go with `env.*`, and you're right that this is likely more intuitive to most users.

But in that case, I'd add to the goals section: discoverability and auditing. By not using the `params` workflow, we can't do things like `ks params list` and `ks params diff`. It seems to me those are valuable, though, so it might be worth special casing them as part of the params workflow.

design/proposals/ambient-namespaces.md
hausdorff: Ok, that sounds good.

design/proposals/ambient-namespaces.md
hausdorff: Yep, that makes sense.

design/proposals/ambient-namespaces.md
jessicayuen: Makes sense. I'll add that section along with an alternatives that details the params approach.

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Proposal
+
+### Fundamentals
+
+The change will behave similarly to how parameters operate in component and prototype manifests.
+
+Components created on the `generate` command will introduce the following line to the top of the component manifest [1]:
+
+`local env = std.extVar("__ksonnet/env");`
```
hausdorff: I'm not crazy about adding more of this sort of thing, but I don't see a better alternative right now, so I'm ok with it. I would be tempted to say to put this in `params.env.*`, but I think your point that these are not mutable, _etc_., is a good one. So I don't know. Anyone else have ideas?

design/proposals/ambient-namespaces.md
```diff
@@ -0,0 +1,126 @@
...
+local appName = "nginx-app";
+
+k.core.v1.list.new([
+  nginx.parts.deployment.withServerBlock(import 'env://namespace', appName),
+  nginx.parts.service(import 'env://namespace', appName),
+  nginx.parts.serverBlockConfigMap(import 'env://namespace', appName),
+])
+```
+
+ksonnet would translate `import 'env://namespace'` prototype snippets to `env.namespace` during the `generate` command.
```
hausdorff: Cool. Just so I understand: the presence of `env://namespace` in a prototype implies an optional flag, `--namespace`, and if that flag is not set, it points at `env.namespace`. Is that right? If so, that makes sense, and we probably want to say that explicitly.

(BTW, I suggest adding the flag only if `env://namespace` is present, because having an optional parameter that does nothing seems odd.)

hausdorff:
I don't think we have a protocol for how to accept design docs, so I voted approve to indicate that I think what's here is pretty close to good enough to start.

hausdorff:
Oh, might want to rename the file to be something other than `ambient-namespaces.md`, though. :)

design/proposals/ambient-namespaces.md
jessicayuen: Hmm, I think my initial intention was to disallow users from being able to set environment metadata through the `generate` command altogether. Ex: Disallowing things like `ks generate some-prototype --namespace=foo`. That way it forces the namespace to be determined during `apply` and what not.

I chose the syntax `import 'env://namespace'` to follow the current custom conventions used by `import 'param://foo'` but there is no reason it can't directly be `env.namespace` (although this might not compile? I'm not sure).

What are your thoughts?

Issue #264, opened 2018-01-04
use /swagger.json from apiserver for api-spec


redbaron:
`ks init` already has access to $KUBECONFIG, it can use it to download /swagger.json to generate metadata.

So it should do something along the lines:
```
kubectl get  --raw /swagger.json | jq '.info += {version: .info.version|sub("\\d+$";"0")}' > /tmp/swagger.json
ks init myapp --api-spec=file:/tmp/swagger.json
```

bryanl:
This is an in-progress idea. We are currently reworking generating the k8s.libsonnet file and I thought this made a lot of sense.

hausdorff:
This was enabled at one point, and then it was disabled for reasons I forget. It's pretty simple to do in itself, we already retrieve the swagger.json for `ks validate`, and the gen code should "just work".

redbaron:
Looking at the same thing from another angle: if swagger.json is provided, no $KUBECONFIG should be needed.

hausdorff:
@redbaron Completely agree.

Issue #232, opened 2017-12-15
Support local folders for registries


jlewi:
I think supporting local folders for registries would be useful for developing packages.

When developing a package, I'd like to be able to try it out without having to push changes to git and reinstalling the package.

I was able to achieve this by creating symbolic links in ./vendor and manually modifying app.yaml

petr-k:
Is using just a plain `git` protocol possible? My use case involves referencing a GitLab repository instead of GitHub.

redbaron:
@petr-k , I am in the same boat. You can drop whole registry management and just checkout package into `vendor` directory, works perfectly well.

ramtej:
Same here. Local folder and/or plan git will be nice!

bryanl:
There will be support for local filesystems in `0.10`. You will be able to `ks registry add <dir>`. This won't as portable as the GitHub based registry, but will be great for local development or combined with the new configuration override functionality.

PR #280, opened 2018-01-09
Design: propose improvements to the "fresh clone" story

hausdorff:
Signed-off-by: Alex Clemmer <clemmer.alexander@gmail.com>

### DIFF
--- a/design/proposals/ks-registry-update-delete.md
+++ b/design/proposals/ks-registry-update-delete.md
@@ -1,0 +1,121 @@
+# `ks registry update` and `ks registry delete`
+
+Status: Pending
+
+Version: Alpha
+
+## Abstract
+
+This proposal describes `ks registry update` and `ks registry delete`, commands
+that will allow users to update and delete the local cache of their registry.
+This is intended to provide a clear path for users to obtain new libraries that
+have been added to the source registry, as well as new versions of old libraries
+in a registry.
+
+## Motivation
+
+The `ks` client tool currently allows user to manage _registries_—collections of
+_packages_ that can be installed and used in the local ksonnet application.
+
+Managing the local cache of a registry when the two diverge is a major painpoint
+in the `ks` client tool. In particular, `ks` currently does not have the ability
+either to update the registry, or to delete a registry, so when a new package is
+added to a remote registry, in order to obtain it, they must:
+
+1. figure out how registries work (_i.e._, via the registry specification in a
+   YAML file), and
+2. manually update their registry spec files (and possibly script it, as [this
+   user has done])
+
+Since users are now creating their own custom registries, we can't afford for
+this part of the workflow to be painful.
+
+The related issue is captured in [#237][237].
+
+## Goal
+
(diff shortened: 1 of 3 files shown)


design/proposals/ks-registry-update-delete.md
```diff
@@ -0,0 +1,121 @@
...
+ks registry update <registry-name> [flags]
+
+Attempt to update our local cache of a package registry. This might include
+adding or removing packages from the cached registry, as well as adding or
+removing specific versions of a package. These packages can be installed into
+the current application with `ks pkg`.
+
+This command additionally provides sanity checks, and will fail in cases where
+updating would make the application un-buildable. For example, if 'registry
+update' would remove a package (or a version of a package) that the application
```
jessicayuen: To make registries easily upgradable, it appears registries should be structured in a way where its packages are version-able. While this may seem obvious, it's less clear when looking at the example registry [incubator](https://github.com/ksonnet/parts/tree/master/incubator) where there's no evidence of versioning for `mysql`, `redis`, etc.

Maybe not as a part of this specific proposal, but as a related goal we should think about how to structure packages in a way that is kind to versioning.  If this functionality already exists, there probably needs to be some documentation so it is clear to registry creators.

hausdorff: Actually packages are versioned! See [here][version]. The version for git registries is a git refspec, in this case `master`. This was hastily done, and we should definitely think more about it. I'll file an issue to follow up.

[version]: https://github.com/ksonnet/parts/blob/master/incubator/registry.yaml#L5

hausdorff: @jessicayuen actually, now that I think about it, I'm ok with our versioning semantics. I propose we actually should wrap the discussion about the versioning semantics into #232, which is when we'd generalize the current registry spec anyway.

hausdorff:
@jessicayuen, @bryanl, there are two proposals here, but they are meant to accomplish the same thing: to improve the "fresh clone" story. Let me know what you think.

bryanl:
The only gotcha I can think of (and I don't even think it is wholly related to this proposal) is in many corporate environments , it will not be possible to reach out to the internet to grab a copy of ksonnet-lib. How would you solve that?

hausdorff:
@bryanl We have a couple options. One is already implemented (and documented in `ks help init`):

```
$ ks init app-name --api-spec=file:swagger.json
```

And the other, not currently enabled (see #264), is to allow users to generate from the `swagger.json` exposed by some API server. Wiring both of these up is basically a matter of:

1. Wiring up the CLI to parse a new API spec type, something like `--api-spec=server:http://example.com`.
1. Wiring up the `ClusterSpec` data structure to correctly retrieve the `swagger.json` from the API server.

I think this is easy enough it doesn't impact the deadline to include this in the proposal if we think it's useful.

bryanl:
Seems useful to take a URL that could be file or http(s).

hausdorff:
@jessicayuen anything else before we merge and get started?

PR #282, opened 2018-01-10
Implement explicit env metadata

### LINKED ISSUE #222: How can components inherit the namespace from the environment?

jessicayuen:
Partially implements the proposal [here](https://github.com/ksonnet/ksonnet/blob/master/design/proposals/explicit-environment-metadata.md).

Version 2 of prototypes in the parts library that follow the historical model of hard-coded namespaces will be updated subsequently. This change does not break existing prototypes.

Fixes #222

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -399,8 +399,10 @@ func expandEnvCmdObjs(cmd *cobra.Command, env string, components []string, cwd m
 		return nil, err
 	}
 
-	libPath, vendorPath, envLibPath, envComponentPath, envParamsPath := manager.LibPaths(env)
-	expander.FlagJpath = append([]string{string(libPath), string(vendorPath), string(envLibPath)}, expander.FlagJpath...)
+	libPath, vendorPath := manager.LibPaths()
+	metadataPath, mainPath, paramsPath, specPath := manager.EnvPaths(env)
+
+	expander.FlagJpath = append([]string{string(libPath), string(vendorPath), string(metadataPath)}, expander.FlagJpath...)
 
 	componentPaths, err := manager.ComponentPaths()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -411,14 +413,20 @@ func expandEnvCmdObjs(cmd *cobra.Command, env string, components []string, cwd m
...
 	}
-	params := importParams(string(envParamsPath))
-	expander.ExtCodes = append([]string{baseObj, params}, expander.ExtCodes...)
+
+	//
+	// Set up ExtCodes to resolve runtime variables such as the environment namespace.
+	//
+
+	params := importParams(string(paramsPath))
+	spec := importEnv(string(specPath))
```
hausdorff: Ah, so, based on the comments in the design proposal, I assumed we were _not_ exposing the server URI or the env name, and instead exposing just the namespace for now?

jessicayuen: Ah sorry, I think this is the one point that I forgot to address in the proposal. The env `spec.json` doesn't contain the name so we would only be exposing the server URI and the namespace.

If I recall correctly, your reservations around exposing the server URI was the `if server == x else` problem. This is the same anti-pattern with namespace and I think it can be discouraged (as mentioned in our other discussion). I also think this pattern will be a lot less seen if `name` isn't exposed. 

Agree `name` is also a ksonnet CLI specific attribute so it makes little sense to expose at the prototype level but `server` and `namespace` are Kubernetes attributes that can be useful as args to ksonnet-lib APIs.

hausdorff: Ok, that makes sense to me. I'm going to mark as "approve", let's see what @bryanl thinks.

jessicayuen:
@bryanl Let me know if you have any additional thoughts before I merge later tomorrow.

PR #283, opened 2018-01-11
Removing realpath

kris-nova:
Darwin (Mac OS systems) does not ship with `realpath` by default

So this removes the `realpath` reference so `make` works out of the box

### DIFF
--- a/docs/generate/update-generated-docs.sh
+++ b/docs/generate/update-generated-docs.sh
@@ -14,7 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-KSONNET_ROOT=$(realpath $(dirname ${BASH_SOURCE})/../..)
+KSONNET_ROOT=$(cd "$(dirname "$0")/../.."; pwd)
+
 BIN=${KSONNET_ROOT}/_output/bin
 mkdir -p ${BIN}
 go build -o ${BIN}/docs-gen ./docs/generate/ks.go


hausdorff:
Not a shell master, at all, but I believe this should always resolve to `./../..`? I don't know a portable way of resolving a path, so I think this might be our best shot. (Unless we want to write our own `realpath`, which I don't think we do.)

@jessicayuen @bryanl what do you think?

hausdorff:
@kris-nova Could you please `commit --amend --signoff`? Required for the DCO tracker. Also if you could just go ahead and delete the commented-out code, that would be great! Thanks.

hausdorff:
(Ha ha, I meant that to sound less office-space-y than it did. "If you could fill out your TPS report that would be great.")

bryanl:
Why not?

```patch
diff --git a/docs/generate/update-generated-docs.sh b/docs/generate/update-generated-docs.sh
index c0f7d4f..2119dc7 100755
--- a/docs/generate/update-generated-docs.sh
+++ b/docs/generate/update-generated-docs.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.

-KSONNET_ROOT=$(realpath $(dirname ${BASH_SOURCE})/../..)
+KSONNET_ROOT=$(cd "$(dirname "$0")/../.."; pwd)
 BIN=${KSONNET_ROOT}/_output/bin
 mkdir -p ${BIN}
 go build -o ${BIN}/docs-gen ./docs/generate/ks.go
```

kris-nova:
Removed commented out code and signed off on my commits - do we need to change the way we are realizing the path or is this fine?

bryanl:
I'm not sure if the relative paths are an issue, that's why i suggested an addition of cd and pwd.

kris-nova:
Ah - didn't realize your diff was a suggestion. Le squashed as well.

hausdorff:
The problem seems to be that the signoff is indented:

```
Removing realpath

      Darwin operating systems does not ship with `realpath` so using another working variation

      Signed-off-by: Kris Nova <knova@heptio.com>
```

If you can remove that space the merge will become unblocked. (Repot is configured so that admins can't force merge with DCO signoff failure.)

Thanks, @kris-nova! Sorry about the hassle.

kris-nova:
😂

PR #285, opened 2018-01-12
Construct apimachinery version

bryanl:
Using `Version()` as a string leaves a preprocessed version tag in the
ksonnet version. Construct a new version using the `Major` and `Minor`
fields.

Re: #250


hausdorff:
Thanks @bryanl for looking into this!

PR #288, opened 2018-01-16
Add remove component functionality

### LINKED ISSUE #243: Remove components
Component remove `ks component rm <component-name>`

- [ ] Clean up component params
- [ ] Remove component file
- [ ] Should tell the user what it is going to delete

jessicayuen:
Fixes #243 

Introduce `ks component rm <component>`. This functionality will:

1. Delete the component file from components/. This is determined by the
component file name. If there exists two files with the same name but
different file  extensions, we will error out.

2. Remove component references. This currently implies removing all
component referenes in the component/params.libsonnet file and the
environment-specific params.libsonnet files.

### DIFF
--- a/cmd/component.go
+++ b/cmd/component.go
@@ -28,6 +28,9 @@ func init() {
 	RootCmd.AddCommand(componentCmd)
 
 	componentCmd.AddCommand(componentListCmd)
+	componentCmd.AddCommand(componentRmCmd)
+
+	componentRmCmd.PersistentFlags().String(flagComponent, "", "The component to be removed from components/")
 }
 
 var componentCmd = &cobra.Command{
@@ -61,4 +64,26 @@ The ` + "`list`" + ` command displays all known components.
 	Example: `
 # List all components
 ks component list`,
+}
+
+var componentRmCmd = &cobra.Command{
+	Use:   "rm <component-name>",
+	Short: "Delete a component from the ksonnet application",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 1 {
+			return fmt.Errorf("'component rm' takes a single argument, that is the name of the component")
+		}
+
+		component := args[0]
+
+		c := kubecfg.NewComponentRmCmd(component)
+		return c.Run()
+	},
+	Long: `Delete a component from the ksonnet application. This is equivalent to deleting the
+component file in the components directory and cleaning up all component
+references throughout the project.`,
+	Example: `# Remove the component 'guestbook'. This is equivalent to deleting guestbook.jsonnet
+# in the components directory, and cleaning up references to the component
+# throughout the ksonnet application.
+ks component rm guestbook`,
 }
--- a/metadata/component.go
+++ b/metadata/component.go
@@ -94,3 +94,107 @@ func (m *manager) CreateComponent(name string, text string, params param.Params,
 	log.Debugf("Writing component parameters at '%s/%s", componentsDir, name)
 	return m.writeComponentParams(name, params)
 }
+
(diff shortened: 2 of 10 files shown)


metadata/component.go
```diff
@@ -94,3 +94,87 @@ func (m *manager) CreateComponent(name string, text string, params param.Params,
...
+
+func (m *manager) DeleteComponent(name string) error {
+	componentPath, err := m.findComponentPath(name)
+	if err != nil {
+		return err
+	}
+
+	// Delete the component file in components/
+	log.Infof("Deleting component '%s' at path '%s'", name, componentPath)
+	m.appFS.Remove(componentPath)
```
hausdorff: I think we should check the error this returns.

metadata/component.go
```diff
@@ -94,3 +94,87 @@ func (m *manager) CreateComponent(name string, text string, params param.Params,
...
+	log.Debugf("... deleting references in %s", m.componentParamsPath)
+	componentParamsFile, err := afero.ReadFile(m.appFS, string(m.componentParamsPath))
+	if err != nil {
+		return err
+	}
+	jsonnet, err := param.DeleteComponent(name, string(componentParamsFile))
+	if err != nil {
+		return err
+	}
+	err = afero.WriteFile(m.appFS, string(m.componentParamsPath), []byte(jsonnet), defaultFilePermissions)
```
hausdorff: I think we should move all FS operations to the very end of the function. Consider a scenario like: we succeed deleting this component, but then something goes wrong as we're parsing the env files, which is not unlikely given the limitations of our transformer at this point. So we abort. This leaves the component half-deleted, and their app in a generally unusable state.

So, overall I'd rather (1) perform all transformations to data, then (2) write out the files and remove things from the FS.

Thoughts?

metadata/component.go
```diff
@@ -94,3 +94,87 @@ func (m *manager) CreateComponent(name string, text string, params param.Params,
...
+		return err
+	}
+
+	// Removing the component references also imply removing the component
+	// references in each environment's environment/<env>/params.libsonnet.
+	envs, err := m.GetEnvironments()
+	if err != nil {
+		return err
+	}
+	for _, env := range envs {
```
hausdorff: Ok, my understanding is that we're not actually removing all references to the component, right? Just the entry in the env for the component.

I think that's fine, but we should adjust documentation to note this caveat, and we should also probably make a TODO to call `lint` after we're finished so that we can tell people when they're breaking something.

metadata/component.go
jessicayuen: This makes sense.

metadata/component.go
jessicayuen: We're removing:
1. The `component/<foo>.jsonnet` file.
2. The references in the `components/params.libsonnet` file.
3. The references in the `environments/<env>/params.libsonnet` file.

What we're not removing and I'll make TODOs for is:
1. References in `main.jsonnet` (mainly because the `main.jsonnet` story is not well scoped and may be subject to change)
2. Component references in other component files, which don't yet exist.

cmd/component.go
```diff
@@ -62,3 +65,25 @@ The ` + "`list`" + ` command displays all known components.
...
+		}
+
+		component := args[0]
+
+		c := kubecfg.NewComponentRmCmd(component)
+		return c.Run()
+	},
+	Long: `Delete a component from the ksonnet application. This is equivalent to deleting the
+component file in the components directory and cleaning up all component
+references thoughout the project.`,
```
bryanl: typo, should be `throughout`

jessicayuen: oops, good catch!

PR #289, opened 2018-01-17
Versions retrospective and fixes

hausdorff:


design/retrospectives/versioning.md
```diff
@@ -0,0 +1,74 @@
...
+* Package versions are not enforced or checked for sanity. And indeed, when we
+  `pkg install`, we actually write down the wrong version when we add the
+  dependency to `app.yaml`.
+
+## Proposed solutions
+
+For 0.8.1:
+
+* Check and enforce all API versions.
+* Create a tag in `ksonnet/parts` called `incubator-v0.1.0`, and hae `ks` 0.8.1
```
jessicayuen: typo: s/hae/have

design/retrospectives/versioning.md
```diff
@@ -0,0 +1,74 @@
...
+  use the version of the registry at that tag by default.
+
+For 0.9.0:
+
+* Since we're pre-1.0, we should break `master` immediately, and simply ask
```
jessicayuen: Is the proposal then to never have a `master` branch?

I'm not sure if this is a good idea to purposefully break `master`. This will require customer integrations like kubeflow to also have to release an updated version.

prototype/prototype_test.go
```diff
@@ -147,3 +149,45 @@ func TestSearch(t *testing.T) {
 	})
 	assertSearch(t, idx, Substring, "foo", []string{})
 }
+
+func TestApiVersionValidate(t *testing.T) {
```
bryanl: Is this the same test that was repeated above? (I think twice) Would it be possible to consolidate this to one place place?

hausdorff: They could, the reason I didn't is: (1) these versions will proceed independently, so the ROI for sharing the test code seems small, and (2) share the test code seems awkward -- in which package should it go? We could re-arrange the way the packages are organized, but I am not sure that this is worth the ROI either.

DO you have a specific idea in mind you'd like to try?

design/retrospectives/versioning.md
hausdorff: We should be depending on nominal versions (_e.g._, `v0.1-beta` or something) rather than `master`. `master` isn't a version name -- git is an implementation detail of the registry.

I'd rather just send a PR to kubeflow than continue on as it is now.

Issue #292, opened 2018-01-19
Difficulty handling components unique to environments


jessicayuen:
Configuration files seem to commonly fall into one or more of these patterns:

1. Configurations that are identical across environments (_i.e._, same in
   _prod_, _dev_, and _staging_).
2. Configurations that are nearly identical, however have simple parameters
   that need to be tweaked across environments (_i.e._, _prod_ accessing a
   different _mysql_ server than _dev_) 
3. Configurations that are similar, however, have chunks of code that are
   entirely different between environments
4. Configurations that are distinctively used in a single environment.

ksonnet does a decent job supporting the first two, however, we currently don't
support 3 and 4 easily, without manipulating the configuration files substantially.

========

Based on the proposals of #295, we'll need to implement:

- [x] *#315* Common metadata handling
- [x] *#309, #308* `app.yaml` to handle environment spec, remove `spec.json`.
- [x] *#319* Add concept of targets and component hierarchy.
- [x] *#319* Update handling of params.libsonnet to include component namespaces
- [x] *WON'T DO, `diff` becomes ambiguous on which clusters to diff.* 
  - *#324* Better multi-cluster deployment support to `ks`
  - Add concept of destinations, update commands accordingly.
  - Introduce `env.yaml` for overriding.
- [x] *#327* Add migration for users using < 0.9.0 environments

jessicayuen:
I'm going to work on a proposal for this, but please feel free to share thoughts @hausdorff @bryanl .

jessicayuen:
Based on the proposals of #295, we'll need to implement:

- [ ] Common metadata handling
- [ ] `app.yaml` to handle environment spec, remove `spec.json`.
- [ ] Add concept of targets and component hierarchy.
- [ ] Update handling of params.libsonnet to include component namespaces
- [ ] Add concept of destinations, update commands accordingly.
- [ ] Introduce `env.yaml` for overriding.

abiogenesis-now:
@jessicayuen what are your thoughts on having either `includes` or `excludes` rather than `targets` (which is basically an `includes`)?

As it is, it's possible to work around this by using a "shared" component subdirectory. However, there's usually some sort of tipping point where either `includes` or `excludes` become easier to write, and having both would let users choose what's right for them.

(I've seen this sort of syntax in Ark backups).

jessicayuen:
@abiogenesis-now Interesting idea, what is the default where an environment doesn't have `includes` or `excludes` specified?

[...]

PR #295, opened 2018-01-23
Proposal: Modular components and cleaner environments

jessicayuen:
Proposal to solve the problems mentioned in #292.

In summary:
* Introduce the concept of _targets_, easily allowing an environment to target specific components.
* Add optional hierarchy structure to components, for grouping and reusability.
* Allow environments to operate on multiple _destinations_.
* Consolidate shared environment metadata and configuration files.

### DIFF
--- a/design/proposals/modular-components.md
+++ b/design/proposals/modular-components.md
@@ -1,0 +1,251 @@
+# Modular Components and Cleaner Environments
+
+Status: Pending
+
+Version: Alpha
+
+## Abstract
+
+This document proposes to provide environments with an easier way of being
+selective of which components to interact with. It also proposes to make
+environments simpler.
+
+## Motivation
+
+The motivation behind this proposal is my user experience while trying to
+onboard Heptio's internal infrastructure configuration files onto ksonnet.
+
+The configuration files seem to commonly fall into one or more of these
+patterns:
+
+1. Configurations that are identical across environments (_i.e._, same in
+   _prod_, _dev_, and _staging_).
+2. Configurations that are nearly identical, however have simple parameters
+   that need to be tweaked across environments (_i.e._, _prod_ accessing a
+   different _mysql_ server than _dev_) 
+3. Configurations that are similar, however, have large pieces of code that
+   differ between environments and are not easily customizable by parameters.
+4. Configurations that are distinctively used in a single environment.
+
+ksonnet currently satisfies the use cases _1_ and _2_, however fall short on
+_3_ and _4_.
+
+## Goal
+
+* Redundancy between component manifests should be kept minimal.
+* It should be simple for an environment to only use a subset of all component
+  manifests.
+* It should be trivial to _add_ or _remove_ components from an environment.
+* It should be trivial to _clone_ an environment.
+* Environments should store the minimal amount of data necessary to operate
+  a set of _components_ on a _k8s cluster_.
+* The 4 use cases mentioned in _Motivation_ should be adequately satisfied.
+
+## Existing State
+
+To understand the remaining portions of this proposal, it is important to have
+context of how environments and components are existingly structured.
+
+This is the current directory structure of a typical ksonnet app:
+
+```text
+my-app
+├── .ksonnet
+├── app.yaml
+├── components
+│   ├── guestbook.jsonnet
+│   ├── params.libsonnet
+│   └── redis.jsonnet
+├── environments
+│   ├── base.libsonnet
+│   ├── default
+│   │   ├── .metadata
+│   │   │   ├── k.libsonnet
+│   │   │   ├── k8s.libsonnet
+│   │   │   └── swagger.json
+│   │   ├── main.jsonnet
+│   │   ├── params.libsonnet
+│   │   └── spec.json
+│   ├── prod
+│   │   ├── .metadata
+│   │   │   ├── k.libsonnet
+│   │   │   ├── k8s.libsonnet
+│   │   │   └── swagger.json
+│   │   ├── main.jsonnet
+│   │   ├── params.libsonnet
+│   │   └── spec.json
+│   └── staging
+│       ├── .metadata
+│       │   ├── k.libsonnet
+│       │   ├── k8s.libsonnet
+│       │   └── swagger.json
+│       ├── main.jsonnet
+│       ├── params.libsonnet
+│       └── spec.json
+├── lib
+└── vendor
+```
+
+The important points to notice are:
+
+1. _components_ have a flat heirarchy, residing in the `components` directory.
+   Components are shared across all environments.
+2. _environments_ are "registered" if there exists a subdirectory in
+   `environments` with the `spec.json` file. `spec.json` contains environment
+   details, such as the cluster server URI and cluster namespace.
+3. _environments_ each contain their own `.metadata` folder containing details
+   of the _ksonnet-lib_ and _kubernetes_ versions being used by the respective
+   environment.
+4. There is a `base.libsonnet` file which acts as the execution entry point
+   when interacting with the clusters specified by a environment. Each
+   environment also contains a `main.libsonnet` file which acts as the
+   "`base-libsonnet` override" file for each environment; allowing for
+   functionality such as the exclusion of components from the specific
+   environment.
+5. Each environment also contains a `params.libsonnet` file that allows for
+   tweaking of simple component parameters (_i.e._ MySQL host) for each
+   environment.
+
+## Proposal
+
+1. Consolidate environment metadata files.
+   * Each environment's `.metadata` directory currently store 3 files related to
+   the generated `ksonnet-lib`. This is unecessary and also costly as the
+   number of environments grow. We needn't store multiple copies of the same
+   API version on disk. The Kubernetes API version that each environment uses
+   can be recorded in the environment specification file. The metadata files
+   can be locally cached once per k8s API version in `lib`.
+
+2. A component should not be uniquely defined by it's base file name.
+   * The existing implementation implies that there cannot be a `foo.jsonnet`
+   and a `foo.yaml`, since the base file name is shared. Instead, a component
+   should be identified by it's relative file path from the components
+   directory.
+
+3. Handle environment specifications in `app.yaml` and retire `spec.json`.
+   * Multiple files and areas for configuration can cause confusions for users.
+   In order to keep configurations simple, I propose that all ksonnet-specific
+   _app-level_ configurations reside in the `app.yaml`.
+
+   `app.yaml` would add an _environments_ section as follows:
+
+   ```
+   apiVersion: 0.0.1
+   kind: ksonnet.io/app
+   name: config
+   version: 0.0.1
+   registries:
+     incubator:
+       gitVersion:
+         commitSha: ea3408d44c2d8ea4d321364e5533d5c60e74bce0
+         refSpec: master
+       protocol: github
+       uri: github.com/ksonnet/parts/tree/master/incubator
+   environments:
+     dev:
+       k8sVersion: 1.9.0
+       destinations:
+         - server: https://my-dev-server
+           namespace: dev
+       targets:
+         - auth
+         - dev
+     staging:
+       k8sVersion: 1.8.0
+       destinations:
+         - server: https://my-prod-server-1
+           namespace: default
+         - server: https://my-prod-server-2
+           namespace: foo
+       targets:
+         - redis.jsonnet
+         - auth
+         - prod
+   ```
+
+   The _environments_ section operates similarly to the existing `spec.json` with
+   the additions of:
+
+   1. Support for multiple destinations. A _destination_ is a server, namespace
+      pair. The configurations specified by the environment will be deployed to
+      all destinations. Destinations can be overwritten locally; see the next
+      point, _4_, for more details on this topic.
+   2. The introduction of _targets_. _Targets_ are the component targets that an
+      environment wishes to include in a deployment. _Target_ definitions are
+      equivalent to the directory or file name of the component. If no targets
+      are specified, all components are targeted by default. `main.jsonnet` will
+      only operate on an environment's target components, while `base.libsonnet`
+      will continue to operate on all components.
+   3. The introduction of _k8sVersion_, the Kubernetes version the environment
+      server is running on.
+
+   Following the above `app.yaml` example, the users directory tree may look
+   something like the following _(hidden files / directories are excluded)_:
+
+   ```
+   my-ksonnet-app
+   ├── app.yaml
+   ├── components
+   │   ├── auth
+   │   │   ├── ca-secret.jsonnet
+   │   │   └── tls-certificate.jsonnet
+   │   ├── dev
+   │   │   └── memcached.jsonnet
+   │   ├── params.libsonnet
+   │   ├── prod
+   │   │   └── database
+   │   │       ├── db-secret.jsonnet
+   │   │       ├── db-service.yaml
+   │   │       └── db-stateful.jsonnet
+   │   └── redis.jsonnet
+   ├── environments
+   │   ├── base.libsonnet
+   │   ├── dev
+   │   │   ├── main.jsonnet
+   │   │   └── params.libsonnet
+   │   └── prod
+   │       ├── main.jsonnet
+   │       └── params.libsonnet
+   ├── lib
+   └── vendor
+   ```
+
+   _Note_: There is no correlation between the environment _dev_ and the
+   components directory _dev_; similarly, with _prod_. We are only calling it as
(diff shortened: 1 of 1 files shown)


design/proposals/modular-components.md
```diff
@@ -0,0 +1,163 @@
...
+components directory _dev_; similarly, with _prod_. We are only calling it as
+such out of convenience. The user can call the components subdirectory whatever
+they desire, so long as it is specified as a _target_ in the `app.yaml` file
+for that environment.
+
+## User Experience
+
+### Backwards compatibility
+
+This change will be *not* be backwards-compatible due to the significant changes
```
bryanl: It could be possible to make this change backwards compatible if we version `app.yaml`

design/proposals/modular-components.md
```diff
@@ -0,0 +1,163 @@
...
+* It should be simple for an environment to only use a subset of all component
+  manifests.
+* It should be trivial to _add_ or _remove_ components from an environment.
+* It should be trivial to _clone_ an environment.
+* Environments should store the minimal amount of data necessary to operate
+  a set of _components_ on a _k8s cluster_.
+
+## Proposal
+
+1. Retire `main.jsonnet` and `base.libsonnet`.
```
bryanl: I can see why these files may still be need. Corporate compliance rules could say you need a project/owner label on all resources. `environments/base.libsonnet` would be a good place to put these. Also, if you wanted to signify the project environment (because your corp uses some type of CMDB for asset management), you may want custom labels as well.

design/proposals/modular-components.md
```diff
@@ -0,0 +1,163 @@
...
+    gitVersion:
+      commitSha: ea3408d44c2d8ea4d321364e5533d5c60e74bce0
+      refSpec: master
+    protocol: github
+    uri: github.com/ksonnet/parts/tree/master/incubator
+environments:
+  dev:
+    k8sVersion: 1.9.0
+    server: https://my-dev-server.com
+    namespaces:
```
bryanl: I propose we use a single namespace until Kubernetes has a better concept of managing multiple resources belonging to a single app across namespaces. If you want to go to another namespace, create another environment.

bryanl:
I added the beginning of a use case in a [gist](https://gist.github.com/bryanl/28afe1d7c0fbd43d409e5dab70f0ef54). It is not yet complete, but I want to make sure we have a use case to build this functionality around.

design/proposals/modular-components.md
jessicayuen: The backwards compatibility is more of an issue of `<= 0.8.0 ks` expecting an environments directory with a `spec.json` file, `.metadata` folder etc.

jessicayuen:
> I added the beginning of a use case in a gist. It is not yet complete, but I want to make sure we have a use case to build this functionality around.

This is great! I'll add some of these points to the proposal.

design/proposals/modular-components.md
abiogenesis-now: +1 to simplicity

design/proposals/modular-components.md
```diff
@@ -0,0 +1,163 @@
...
+1. Retire `main.jsonnet` and `base.libsonnet`.
+   The purpose of these files are often confusing to users. The goal of 
+   `main.jsonnet` was to act as a way for environments to provide
+   environment-specific extensibility to components. However, this only works
+   if the component is written in jsonnet, and also requires the user to have
+   an advanced understanding of jsonnet. Furthermore, with the existing model,
+   all components are operated on during `apply`, `delete`, etc. If the
+   components vary drastically between environments, it is not convenient to
+   "subtract" these components from each environment's `main.jsonnet`.
+   `base.libsonnet` also currently serves no purpose besides acting as an entry
```
abiogenesis-now: what do you think of removing `main.jsonnet` and `base.libsonnet` but still keeping env metadata within the env directories? 

e.g.  `env.yaml` (?) which includes the targets you mentioned (?) as well as server info

abiogenesis-now: because right now there's this strange split between `<env>/params.jsonnet` and the top-level `app.yaml`

jessicayuen: I don't have too strong of opinions for either approach. I think the benefits of keeping it in `app.yaml` is the ease to see configurations in one place (especially when the configuration isn't very long). And the cons are, this will keep the environments directory quite bare -- need to think more on what purpose the directory structure serves, and how to repurpose `params.libsonnet`, if needed.

design/proposals/modular-components.md
jessicayuen: Yes, the idea is that multiple namespaces should only be specified for an environment if the anticipation is that all components / params across the namespaces are identical. The scenario where multiple resources should be split across namespaces, should be modeled as multiple environments.

design/proposals/modular-components.md
```diff
@@ -0,0 +1,234 @@
...
+   tweaking of simple component parameters (_i.e._ MySQL host) for each
+   environment.
+
+## Proposal
+
+1. Retire `main.jsonnet` and `base.libsonnet`.
+   * The purpose of these files are often confusing to users. The goal of
+   `main.jsonnet` was to act as a way for environments to provide
+   environment-specific extensibility to components. However, this only works
+   if the component is written in jsonnet, and also requires the user to have
```
redbaron: > this only works if the component is written in jsonnet

how else component could be written?

design/proposals/modular-components.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+The _environments_ section operates similarly to the existing `spec.json` with
+the additions of:
+
+1. Support for multiple environment namespaces. Deployments will happen across
+   all the namespaces for that environment. If a user wishes to have separate
+   deployments to separate namespaces, these deployment groups should be
+   modeled as two separate environments.
+2. The introduction of _targets_. _Targets_ are the component targets that an
```
redbaron: Is there a place where all components are available as a single objects (much like now retired `base.libsonnet`?). It was handy to have as CI pipeline could drop arbitrary code there and perform various validations and mandatory transformations (such as adding annotations and labels to ALL objects without need for modifying any components).

design/proposals/modular-components.md
```diff
@@ -0,0 +1,234 @@
...
+    protocol: github
+    uri: github.com/ksonnet/parts/tree/master/incubator
+environments:
+  dev:
+    k8sVersion: 1.9.0
+    server: https://my-dev-server.com
+    namespaces:
+      - dev-foo
+      - dev-bar
+    targets:
```
redbaron: IMHO more often than not, every environment would have same amount of targets. Making users to specify all targets for all envs feels like optimizing for less common case and would cause errors such as new component was specified in `dev`, tested and verifies, then app goes to `prod` and component is missing and app is on fire.

redbaron:
I am not sure if it is the right scope for this proposal, but I'd like problem of component depdendency to be solved somehow. Right now I have "glue" functions I call in `base.libsonnet` which extract params from one component and injects them into another, but this is not sustainable approach as component specific logic is then spread over.

Classic example would be to get service name of one component and use it in another. Helm solves it by convention, which is too fraglile and require whole ecosystem to be aligned with a given convention. Ideally there should be some explicit way  for one component to pull final resulting fields of another component. Amount of coupling here is an open question for debate

design/proposals/modular-components.md
chancez: Unrelated to the above comment, I think the concept of `targets` is hugely important. This is one area where I found ksonnet seemed to be lacking/difficult. It's very common for me in my use-cases (as described in your user stories above) to want to omit deploying specific resources and components typically based on environment.

I do agree with redbaron that being able to basically transform all resources in some way would be nice, I can see that being useful to set ownerRefs/annotations across all objects if an operator was running ksonnet or something.

design/proposals/modular-components.md
chancez: I think supporting many namespaces where everything is uniform is a pretty typical use case. I can imagine one such example being to setup common RBAC policies in every namespace, setting up an ingress controller per namespace, and doing quotas/limitranges. Depending on what's being done, it may not always be the same by namespace, but in some of these cases, it might be.

design/proposals/modular-components.md
chancez: I disagree, this is something we would heavily use where I work.

Feature flagging is a common thing I do today with helm to disable specific parts of the stack being deployed depending on the environment. It's very common to build and add a new microservice to the stack, but it's often not deployed to all environments until it's been fleshed out in dev more.

I think having an explicit list of targets is also hugely important for avoiding accidental changes when a component changes but the `app.yaml` doesn't.  This pattern also exists in many other places like config management, where you really want to be explicit about what's being ran for each environment, in case the upstream adds a ton of new components, and you want to ensure you're only deploying what's needed when updating it.

bryanl: The real world use case we've seen is when dev and production differ because of security requirements. The goal here is to give a good default and allow for more flexibility when it is needed.

chancez: Also, I believe this could be accommodated by defining a component `default.jsonnet` which just effectively imports the other components, and in the `app.yaml`, you would just specify `targets: [ "default.jsonnet" ]`.

design/proposals/modular-components.md
```diff
@@ -0,0 +1,234 @@
...
+├── environments
+│   ├── dev
+│   │   └── params.libsonnet
+│   └── prod
+│       └── params.libsonnet
+├── lib
+└── vendor
+```
+
+_Note_: There is no correlation between the environment _dev_ and the
```
chancez: This might be bike shedding, but I feel like the figuring out best practices around environments/targets is good to earlier than later, and thus may effect the design so I wanted to comment on the example.

While I understand the design doesn't preclude this, it feels like the example here is an anti-pattern. 

I think it would make more sense to put the components at a level up (eg: `components/database` and `components/memcached.jsonnet), and then if you wanted `prod` or `dev`, it would make more sense to have a `dev.jsonnet` and `prod.jsonnet` which imports the env specific components and calls them directly, passing params through.

So then the `staging` environment from the example would adjust it's targets to this:
```
staging:
   targets:
     - prod.jsonnet
```

This wouldn't preclude listing more targets obviously as it was originally, but fits more into the documented model of a component, where a component could be something as complex as an entire environment. It also makes all the individual components more or less environment agnostic (is the db really prod only? is memcached, really dev only?). The alternative is that this `prod.jsonnet` or `dev.jsonnet` could live outside the `components` and be referenced as an alternative target.

Comparing this Chef in config management land, this is basically like `components==recipes`, `environments==environments/app.yaml`, and `roles`. are effectively a the `targets` in app.yaml. The idea in Chef is that environments can refer to individual recipes (components), or roles, and roles provide a way to group various recipes (components) together. I feel like we want a similar concept here, to group components either by role, environment, etc, which can be referred to as a target, as an alternative to listing all targets in app.yaml.

design/proposals/modular-components.md
redbaron: @chancez ,

> It's very common to build and add a new microservice to the stack, but it's often not deployed to all environments until it's been fleshed out in dev more.

Looks like you are using single ksonnet app to describe whole stack, something which would be pulling in tens of components, which vary between environment depending how mature they are. if so is it the right scope for single ksonnet app?  One of the promises of microservices is independent release cycle, instead of one big bang, by pulling them all under single umbrella you deny that kind of flexibility.

Maybe there is a different view of what component is? For me ksonnet app is a single app which provides a single service. It might require some dependencies for it's own operation, but these dependencies are more like implementation detail of that app. Dependencies which are shared in implementation (but not during operation) make sense to factor out as  components, for example: redis, memcache, rabbitmq, etc.   From this POV, there will be very rare change between environments regarding components app is installing: if it requires redis for its operation, it is going to need it in every environment.

redbaron: @chancez ,

> I think having an explicit list of targets is also hugely important for avoiding accidental changes when a component changes but the app.yaml doesn't.

I can't understand which failure scenario you are describing, would you mind to elaborate? Components in ksonnet so far are installed prototypes, which themselve are  small pieces of code calling package they come from in a particular way.  Package versions are in full control of user, they can't accidentally change. All prototypes from a package need be installed explicitly and become component to have any effect on ksonnet app. Simple fact that there is  a component is already explicit enough to avoid surprises.

chancez: I think both ways makes sense. I believe that ideally, you can build tons of individually deployable components, and that should work. But I think being able to deploy them together via ksonnet is also important. Whether you use ksonnet to deploy groups of apps, or apps individually depends on size of teams and various other factors.

Just because you can develop and release microservices individually for example, doesn't mean that's how it will work in all scenarios. You might be deploying and releasing microservices individually in pre-prod, but it's entirely likely for production things are released together.

redbaron: > , but it's entirely likely for production things are released together.

Of course you can choose to release them together, but should it really be done as a single ksonnet app describing whole stack? There are simply no tooling in ksonnet to orchestrate such whole stack rollout as a single app: no dependencies, no inbetween smoke tests, no custom success conditions and notifications.

chancez: I'd argue that's a problem of ksonnet already, today.

Ex, if you're deploying $microService and Redis together, you're already in a situation where you cannot orchestrate the rollout nor the interaction between these two components ,and they're coupled to each other. In dev, who cares, in staging/prod, maybe they should each be deployed individually with things like smoke tests running between each component deployment.

I'm not saying grouping of many resources in a single deployment is the best practice, but it's a very reasonable one depending on your needs. Many systems or applications simply don't need to do anything other than roll out all new things together. I entirely agree if you need to run various tests, and other things between rollouts, you'll likely need tooling on top of ksonnet, and your components will need to be broken out accordingly, but I think having flexibility here is important. 

Having to orchestrate many individual deploys itself is complex, and often it doesn't need to be. People are already fine running `kubectl apply -f` or `helm install` on large groups of resources, without a heavy need for handling each individual component's roll out.

redbaron: There is a big difference, that components of an app are all managed by the same team, where upgrade path is excersised much more often and there are much less communication distance than between teams. Orchestrating update of a whole stack, especially when it done for the first time in prod (as dev is done independently as per discussion above) is  a much biggger task and it is not purely technical. There is also difference in scope: single digit number of compoonents can be coordinated with initcontainers and entrypoints if needed, but more often than not just putting them all there is enough.

Anyway, this thread is not about how companies should manage their CD pipeline, it is about not optimizing for case which ksonnet is less than an optimal tool right now.  I am all for explicit over implicit, but I am also against unnecessary duplication: simple fact that component is there is explicit enough to indicate that ksonnet app needs it.

What about defaulting list of targets to all components  and using explicit `targets:` if specified only? This way spcial environments can list all their components, but more common cases will be covered automatically: all present components are installed.

chancez: I think i'd rather see a feature where could expose different lists of targets. Perhaps a concept like a role. You could require a default role be defined, which would contain the list of targets to use by default if targets isn't explicitly set. This would also then allow different lists of targets to be defined, and used. Of course, you could set targets that list components directly also.

I can definitely see in the long run if the option to specify targets exists (which I think we can agree is useful) then how people write their apps will change. Potentially writing many components, but not expecting all of them to be installed. Having a way to declare the default list of them, and alternatives would be much more flexible. 

In the end, you should be able to define what components to expose for the default installation, and not necessarily all should be used by default.

jessicayuen: What I failed to mention in the proposal is if a target isn't specified, all components will be targeted by default. This is parallel to how ksonnet operates currently, so it should solve both scenarios mentioned in this conversation.

design/proposals/modular-components.md
jessicayuen: We don't necessarily need to retire `base.libsonnet`, especially if a strong use case like this one presents. Targets can operate in parallel.

design/proposals/modular-components.md
jessicayuen: This is great, thanks for all your feedback.

I agree that the example in the proposal might have poorly highlighted how a real scenario might be modeled. I wanted to present that if a production and a dev environment had a large set of configurations that differ greatly from each other (_i.e._ `database` in this scenario), it could be modeled in their own directory. For example,

```
├── prod
│   └── database
│       └── a-config.jsonnet
├── devel
│   └── database
│       └── a-completely-different-config.jsonnet
```

```
prod:
  targets:
    - prod
dev:
  targets:
    - devel
```

If `database` has the ability to be shared easily across multiple environments, then it makes sense to have `database` at the top-level, as you mentioned.

I believe your proposal of `prod.jsonnet` and `dev.jsonnet` is similar to how `main.jsonnet` for each environment operates today. We're still evaluating whether they exist well alongside this _targets_ model.

Issue #11, opened 2017-10-16
Formalize unique cluster ID on env creation 


jessicayuen:
_From @jessicayuen on September 18, 2017 18:59_

> We should get formal about how the env commands produce a unique ID for a cluster. I'm thinking a hash of URI + server cert? This is a temporary solution of course, we'll need to fix it up stream.

_Copied from original issue: ksonnet/kubecfg#135_

jessicayuen:
_From @seh on September 23, 2017 20:17_

In my organization, we've had clusters where we reach the API server via a _Service_ of type "NodePort," such that there are any number of node IP addresses and one specific port that all point to the "same" cluster. It's a mess, because different people can have different server addresses in their KUBECONFIG file, and it's hard to tell whether they're referring to the same cluster. We usually first compare the port, and if those match, advise to converge on a single IP address (picking one node, which is fragile, of course) to make it easier to identify equivalence in the future.

I'm not saying that finding the API server via node port is a good way to operate a cluster. I just want to point out that cluster identity doesn't come from the server address, which is a surjective reference. In short: Yes, this is a hard problem. Relying on the user to give different clusters different names, and not duplicating any clusters (just with different addresses, or with a different CA certificate bundle) is a sane way to deflect the problem.

bryanl:
This functionality was implemented via overrides.

PR #300, opened 2018-01-27
environment id proposal

bryanl:
Re: https://github.com/ksonnet/ksonnet/issues/11


jessicayuen:
My only concern with this is it will force environments to be 1:1 with a namespace. _i.e._ we wouldn't be able to instantiate multiple environment for a namespace. The configurations of an environment will also not be able to be deployed to multiple namespaces. 

Guess this will be directly related to the outcome of the discussions on #295 .

bryanl:
closing this for now

PR #301, opened 2018-01-28
use afero when possible

### LINKED ISSUE #39: Disallowed `ioutil.*` functions used
Currently we use `afero` to mock all FS calls. We're calling some functions we shouldn't, like `ioutil.ReadFile` in various places. We should replace these with `afero` calls.

bryanl:
Use afero when possible to allow for fs flexibitily (in tests or other instances).

fixes #39

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -29,7 +29,9 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
 	"k8s.io/client-go/discovery"
@@ -61,11 +63,16 @@ const (
 	flagComponentShort = "c"
 )
 
-var clientConfig clientcmd.ClientConfig
-var overrides clientcmd.ConfigOverrides
-var loadingRules clientcmd.ClientConfigLoadingRules
+var (
+	clientConfig clientcmd.ClientConfig
+	overrides    clientcmd.ConfigOverrides
+	loadingRules clientcmd.ClientConfigLoadingRules
+	appFs        afero.Fs
+)
 
 func init() {
+	appFs = afero.NewOsFs()
+
 	RootCmd.PersistentFlags().CountP(flagVerbose, "v", "Increase verbosity. May be given multiple times.")
 
 	// The "usual" clientcmd/kubectl flags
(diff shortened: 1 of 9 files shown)


template/expander.go
```diff
@@ -24,8 +28,21 @@ type Expander struct {
 
 	Resolver   string
 	FailAction string
+
+	fs afero.Fs
+}
+
+func NewExpander(fs afero.Fs) Expander {
+	if fs == nil {
+		fs = genDefaultFs()
```
jessicayuen: Since `genDefaultFs()` isn't being used anywhere else, this can probably just be `fs = afero.NewOsFs()`.

cmd/root.go
```diff
@@ -61,11 +62,17 @@ const (
...
 
-var clientConfig clientcmd.ClientConfig
-var overrides clientcmd.ConfigOverrides
-var loadingRules clientcmd.ClientConfigLoadingRules
+var (
+	clientConfig clientcmd.ClientConfig
+	overrides    clientcmd.ConfigOverrides
+	loadingRules clientcmd.ClientConfigLoadingRules
+	appFs        afero.Fs
+	defaultFs    = afero.NewOsFs()
```
jessicayuen: I believe this isn't used, can be deleted.

cmd/root.go
```diff
@@ -384,8 +391,8 @@ func overrideCluster(envName string, clientConfig clientcmd.ClientConfig, overri
 // the user passes a list of files, we will expand all templates in those files,
 // while if a user passes an environment name, we will expand all component
 // files using that environment.
-func expandEnvCmdObjs(cmd *cobra.Command, env string, components []string, cwd metadata.AbsPath) ([]*unstructured.Unstructured, error) {
-	expander, err := newExpander(cmd)
+func expandEnvCmdObjs(fs afero.Fs, cmd *cobra.Command, env string, components []string, cwd metadata.AbsPath) ([]*unstructured.Unstructured, error) {
```
jessicayuen: Instead of passing `fs` from every command, I think the below can just be `expander, err := newExpander(appFs, cmd)`. This should simplify this change a lot and remove most of the `fs` param passing.

bryanl: The reason I did this is so we aren't reliant on global state. I think I'll take another stab at rewriting this so the process can rely on global state as a default, but still allow you to pass a fs in as an override.

cmd/root.go
```diff
@@ -379,39 +386,64 @@ func overrideCluster(envName string, clientConfig clientcmd.ClientConfig, overri
...
-// expandEnvCmdObjs finds and expands templates for the family of commands of
+type cmdObjExpanderConfig struct {
+	fs         afero.Fs
+	cmd        *cobra.Command
+	env        string
+	components []string
+	cwd        metadata.AbsPath
+}
+
+// templateExpands finds and expands templates for the family of commands of
```
jessicayuen: s/templateExpands/cmdObjExpander

PR #308, opened 2018-02-01
Add interface for Environment Spec

jessicayuen:
- Add EnvironmentSpec to app.yaml
- Move the component param logic from metadata/manager.go into metadata/component.go

Partially implements #292

### DIFF
--- a/metadata/app/schema.go
+++ b/metadata/app/schema.go
@@ -29,8 +29,14 @@ const (
 	DefaultVersion    = "0.0.1"
 )
 
-var ErrRegistryNameInvalid = fmt.Errorf("Registry name is invalid")
-var ErrRegistryExists = fmt.Errorf("Registry with name already exists")
+var (
+	ErrRegistryNameInvalid = fmt.Errorf("Registry name is invalid")
+	ErrRegistryExists      = fmt.Errorf("Registry with name already exists")
+	// ErrEnvironmentNameInvalid is the error where an environment name is invalid.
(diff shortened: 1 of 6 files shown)


metadata/app/schema.go
```diff
@@ -29,8 +29,12 @@ const (
 	DefaultVersion    = "0.0.1"
 )
 
-var ErrRegistryNameInvalid = fmt.Errorf("Registry name is invalid")
-var ErrRegistryExists = fmt.Errorf("Registry with name already exists")
+var (
+	ErrRegistryNameInvalid    = fmt.Errorf("Registry name is invalid")
```
bryanl: golint will complain because these exported items don't have a comment.

metadata/component.go
```diff
@@ -171,6 +171,38 @@ func (m *manager) DeleteComponent(name string) error {
 	return nil
 }
 
+func (m *manager) GetComponentParams(component string) (param.Params, error) {
```
bryanl: `manager` has a large number of functions. As you change these, it's a good time to start extracting functionality into other structures.

metadata/app/schema.go
jessicayuen: This is existing code from the Registry portions of the schema. I've added comments on all new exported Environment Spec related code. Since registries are being re-written I didn't want to spend time to fix existing lint errors related to registries, and wanted to keep this change contained to Environments / Components.

PR #309, opened 2018-02-02
Migrate environment spec.json to the app.yaml model

jessicayuen:
Partially implements #292 

Currently spec.json contains detail about an environment's namespace and
server. Following the proposal at
design/proposals/modular-components.md, this change will consolidate
environment specifications in the common ksonnet app.yaml file.

An environment specification for the environment "dev", may look as
follows:

```
environments:
  dev:
    destinations:
    - namespace: foo
      server: example.com
    k8sVersion: "1.8.1"
    path: dev
    targets:
    - db
```

Note: This change currently doesn't support

(1) population of the k8sVersion field. This will occur as we migrate
the environment .metadata folder.
(2) deployment to more than one destination. This will occur once ks
supports multi-cluster deployment.
(3) setting of details other than the env name in `ks env set`. Prior
to this change, users are able to namespace and server URI, however it
becomes ambiguous which namespace is being set for an environment where
there can be multiple destinations. We will encourage configuration in
app.yaml itself.
(4) targets. This will come in a later change.


jessicayuen:
@bryanl The failure scenario for those using `spec.json` is the environment not being registered, i.e.: You might get the following error,

```
/ks apply default
ERROR Environment 'default' does not exist
```

The migration path would involve adding those details to the `app.yaml`. This can be done manually, but a migration path could also be built in. Did you have suggestions for the entry point of the automated migration?

bryanl:
Yes, we can temporarily say that a change needs to happen, but we should offer to upgrade before the 0.9.0 release.

PR #315, opened 2018-02-07
Add LibManager for managing k8s API and ksonnet-lib metadata

jessicayuen:
This change will introduce a lib package that handles the reading,
writing, and serialization of the ksonnet-lib and the kubernete's open
API files.

Prior to this change, metadata was stored in an environment's .metadata
directory. This lead to redundant files where environments share the
same kubernetes API version.


jessicayuen:
@bryanl Just to clarify, so that I understand the change you are proposing - I didn't create the `utils` package; it exists already. I just moved `AbsPath` out of metadata into the existing utils package, since it is useful for packages outside of the app metadata. Where are you suggesting that it go?

I can't speak fully for the reason `string` is typed as `AbsPath` as I didn't write that, but I believe that the logic behind it is the path format is too important to be an arbitrary string. It makes it more obvious to the developers of the codebase that we are in fact dealing with a path.

jessicayuen:
@bryanl, I've made the changes we discussed offline:

1. Moving utils/strings.go > strings/strings.go
2. Converting AbsPath to simple strings and adding function `strings.AppendToPath`

PR #323, opened 2018-02-15
Introduce component namespaces

bryanl:
Introduces component namespaces and support for targets in `app.yaml`

* Adds support for targets in `app.yaml`. You can now specify targets for an environment so it only operates on a subset of components. (or you can not define it and it will operate on all components)
* Extracts the core of code use to manage components to a new package `github.com/ksonnet/ksonnet/comonent`. 
* Introduces component namespaces. Components can now be broken up into separate namespaces. Combined with targets, you can apply the exact subset of components needed for an environment.

Commands to be fixed

- [x] `ks component list`
- [x] `ks component rm`

re: #292


bryanl:
I'll move the the other component flows as well to make the idea complete.

component/component_test.go
```diff
@@ -0,0 +1,426 @@
...
+		"/components/params.libsonnet",
+		"/components/nested/a.jsonnet",
+		"/components/nested/params.libsonnet",
+		"/components/nested/very/deeply/c.jsonnet",
+		"/components/nested/very/deeply/params.libsonnet",
+		"/components/shallow/c.jsonnet",
+		"/components/shallow/params.libsonnet",
+	}
+
+	invalidPaths = []string{
```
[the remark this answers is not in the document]
bryanl: I changed the file extension. What we really need is a better way to identify files.

cmd/root.go
```diff
@@ -433,40 +434,54 @@ func (te *cmdObjExpander) Expand() ([]*unstructured.Unstructured, error) {
...
 	}
 
-	expander.ExtCodes = append([]string{baseObj, params, envSpec}, expander.ExtCodes...)
+	baseCodes := expander.ExtCodes
 
-	//
-	// Expand the ksonnet app as rendered for environment `env`.
-	//
+	slUnstructured := make([]*unstructured.Unstructured, 0)
+	for ns, componentPaths := range namespacedComponentPaths {
```
jessicayuen: The library paths are being repeatedly added because of this loop.

_i.e._ on _apply_, for every component namespace, we will be getting same sets of these messages:
```
DEBUG Adding jsonnet search path /Users/jyuen/workspace/go/src/github.com/ksonnet/ksonnet/foo/vendor
DEBUG Adding jsonnet search path /Users/jyuen/workspace/go/src/github.com/ksonnet/ksonnet/foo/lib/v1.7.0
```

Can you change this so that the lib paths are added only once?

bryanl: To clear this up, they are only added once for each invocation. The issue is that the `jsonnetVM` code has to run for each namespace separately. There way to solve it is to have the expander worker with multiple component namespaces and that might be more trouble than it is worth to get this functionality started.

bryanl: For posterity, here is what is generated if you filter out stderr.

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: guiroot
spec:
  ports:
  - port: 80
    targetPort: 80
  selector:
    app: guiroot
  type: ClusterIP
---
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: guiroot
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: guiroot
    spec:
      containers:
      - image: gcr.io/heptio-images/ks-guestbook-demo:0.1
        name: guiroot
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: gui
spec:
  ports:
  - port: 80
    targetPort: 80
  selector:
    app: gui
  type: ClusterIP
---
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: gui
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: gui
    spec:
      containers:
      - image: gcr.io/heptio-images/ks-guestbook-demo:0.1
        name: gui
        ports:
        - containerPort: 80
```

PR #327, opened 2018-02-22
Add warning for running deprecated ks app against ks >= 0.9.0

jessicayuen:
This is a temporary piece of code that should be removed after 1.0.0.

This warning exists to tell users that are using ks < 0.9.0 with the
dated environments model (contains the 'spec.json' file), to migrate to
the newer 'app.yaml' model.

### DIFF
--- a/metadata/environment.go
+++ b/metadata/environment.go
@@ -385,6 +385,34 @@ func (m *manager) EnvPaths(env string) (libPath, mainPath, paramsPath string, er
 	return
 }
 
+func (m *manager) ErrorOnSpecFile() error {
+	return afero.Walk(m.appFS, m.environmentsPath, func(p string, f os.FileInfo, err error) error {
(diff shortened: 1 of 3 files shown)


metadata/environment.go
```diff
@@ -385,6 +385,34 @@ func (m *manager) EnvPaths(env string) (libPath, mainPath, paramsPath string, er
...
+			log.Debugf("Failed to walk path %s", p)
+			return err
+		}
+		isDir, err := afero.IsDir(m.appFS, p)
+		if err != nil {
+			log.Debugf("Failed to check whether the path at %s is a directory", p)
+			return err
+		}
+		if isDir {
+			specPath := path.Join(p, "spec.json")
```
bryanl: This app can potentially run on Windows so we should be using `filepath.Join`

PR #328, opened 2018-02-23
App spec to take a single destination

jessicayuen:
Currently the app.yaml spec takes a list of destinations. The change
exists to anticipate support for an environment supporting multiple
clusters or "destinations". The problem is, while the use case makes
sense for `apply`, and `delete`, it becomes ambiguous which cluster is
being referred to during commands such as `diff`. We've considered
specifying a cluster during a `diff`, however, ks currently doesn't have
a notion of a cluster identity. This change is to update the app.yaml to
take a single destination to more accurately represent the state of
things.

### DIFF
--- a/metadata/app/schema.go
+++ b/metadata/app/schema.go
@@ -75,33 +75,29 @@ type RegistryRefSpecs map[string]*RegistryRefSpec
 type EnvironmentSpecs map[string]*EnvironmentSpec
 
 // EnvironmentSpec contains the specification for ksonnet environments.
-//
-// KubernetesVersion: The Kubernetes version the target cluster is running on.
-// Path:              The relative path containing metadata for this environment,
-//                    rooted at the directory 'environments'.
-// Destinations:      One or more cluster addresses that this environment
-//                    points to.
-// Targets:           The relative component paths that this environment wishes
-//                    to deploy onto it's destinations.
 type EnvironmentSpec struct {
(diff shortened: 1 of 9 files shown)


bryanl:
If we are removing the ability to handle multiple destinations, how easy would it be to add it back in the future? I'd hate to put us in a place where it is very difficult to handle multiple clusters.

jessicayuen:
@bryanl This change won't be removing the ability to handle multiple destinations (since we don't have that code yet!). It's just removing the ability to specify multiple destinations for a environment in the `app.yaml`. Because if we decide to not support applying or deleting to multiple destinations (for the `diff` reasons), it might confuse users on why they can specify multiple destinations in the `app.yaml`.

metadata/app/schema.go
```diff
@@ -79,23 +79,19 @@ type EnvironmentSpecs map[string]*EnvironmentSpec
 // KubernetesVersion: The Kubernetes version the target cluster is running on.
 // Path:              The relative path containing metadata for this environment,
 //                    rooted at the directory 'environments'.
-// Destinations:      One or more cluster addresses that this environment
-//                    points to.
+// Destination:       Cluster address that this environment points to.
```
bryanl: These comments should be in the type definition so they show up properly with godoc. Currently there will be docs on the types, but not the members. The same thing applies to `EnvironmentDestinationSpec` as well.

PR #330, opened 2018-02-25
Create support for plugins in ksonnet

bryanl:
This will be a multi-phased approach. In the first phase, allow ksonnet to run plugins. These plugins will allow for faster iteration on new and improved functionality without impacting the core of the application. 
ks-plugins-2


Gopkg.toml
```diff
@@ -114,4 +114,7 @@
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "5.0.0"
\ No newline at end of file
+  version = "5.0.0"
+[[constraint]]
+  branch = "master"
+  name = "github.com/shibukawa/configdir"
```
jessicayuen: Don't believe that this dependency is being used

cmd/root.go
```diff
@@ -109,6 +110,57 @@ application configuration to remote clusters.
...
+		p, err := plugin.Find(appFs, pluginName)
+		if err != nil {
+			return err
+		}
+
+		return runPlugin(p, args)
+	},
+}
+
+func runPlugin(p plugin.Plugin, args []string) error {
```
jessicayuen: What's your opinion on putting the this plugin function to the new `plugin` package?

bryanl: It does belong there, but the information I wanted to share isn't easily accessible from there. Need to go back and make that easier to retrieve. Then it can move.

PR #337, opened 2018-03-01
Attempt to generate lib directory when not found

bryanl:


bryanl:
@jessicayuen I believe I fixed so you get the privilege of approving your own PR :)

PR #338, opened 2018-03-02
ksonnet app.yaml format changes in next minor release. Handle both versions

bryanl:
* Environment can support 0.1.0 and 0.0.1 formats
* Provide command to upgrade older formats to 0.1.0


cmd/upgrade.go
```diff
@@ -0,0 +1,61 @@
...
+The upgrade command upgrades a ksonnet application to the latest version.
+
+### Syntax
+
+	Example:
+
+# Upgrade ksonnet application in dry-run mode to see the changes
+ks upgrade --dry-run
+
+# Upgrade ksonnet application. This will update app.yaml to apiVersion 0.1.0
```
jessicayuen: Not sure if you want to include version detail and specifics in here, in the event of using this command for future ks versions. Might be just worth saying something along the lines of: 
```
Upgrade ksonnet application to the version latest defined by _ks_.
```

generator/ksonnet.go
```diff
@@ -44,9 +60,16 @@ func Ksonnet(swaggerData []byte) (*KsonnetLib, error) {
 		return nil, err
 	}
 
-	spew.Dump("---", f.Name(), ksonnetEmitter)
+	astMin := semver.MustParse("1.8.0")
```
jessicayuen: Is this the minimum version supported by ks? If yes, what about 1.7.0?

metadata/interface.go
```diff
@@ -54,19 +55,18 @@ type Manager interface {
...
-	GetEnvironments() (app.EnvironmentSpecs, error)
-	GetEnvironment(name string) (*app.EnvironmentSpec, error)
+	GetEnvironments() (map[string]env.Env, error)
+	GetEnvironment(name string) (*env.Env, error)
 	SetEnvironment(name, desiredName string) error
 
-	// Spec API.
-	AppSpec() (*app.Spec, error)
-	WriteAppSpec(*app.Spec) error
+	// Destination API.
```
jessicayuen: I would just bundle this with the `// Environment API` group since destinations are a part of environments.

generator/ksonnet.go
bryanl: No, it will use the older method of generating ksonnet-lib if you are below 1.8.0. It still works with 1.7.0+.

Gopkg.lock
```diff
@@ -255,6 +255,18 @@
   packages = ["."]
   revision = "fc9e8d8ef48496124e79ae0df75490096eccf6fe"
 
+[[projects]]
+  name = "github.com/mattn/go-runewidth"
```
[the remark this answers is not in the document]
bryanl: Go doesn't have test dependencies. They are being using by testify mocks.

Gopkg.lock
```diff
@@ -359,10 +371,17 @@
   packages = ["."]
   revision = "e57e3eeb33f795204c1ca35f56c44f83227c6e66"
 
+[[projects]]
+  name = "github.com/stretchr/objx"
```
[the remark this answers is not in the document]
bryanl: As above ☝🏿

PR #379, opened 2018-03-21
Create file system based part

bryanl:
* add filesystem based registry
* relocate part related code to `pkg/registry`


bryanl:
Fixes #232

PR #462, opened 2018-04-20
When importing YAML, extract objects into separate component files

### LINKED ISSUE #437: ks import -f some-component.yaml does not set component index in components/params.libsonnet
**Is this a BUG REPORT or FEATURE REQUEST?**: BUG REPORT
/kind bug

**What happened**:

After importing a yaml using `ks import`, the params.libsonnet is missing the -X index.
```
$ ks import -f ~/tmp/kube-ops-view-deploy.yaml
INFO Writing component at '/Users/jsuen/ws/myapps/kube-ops-view/components/kube-ops-view-deploy.yaml'

$ cat components/params.libsonnet
{
  global: {
  },
  components: {
    // Component-level parameters, defined initially from 'ks prototype use ...'
    // Each object below should correspond to a component in the components/ directory
    "kube-ops-view-deploy": {

    },
  },
}
```

According to chat with @bryanl, the component name in params.libsonnet should be: "kube-ops-view-deploy-0" instead of "kube-ops-view-deploy"

**How to reproduce it (as minimally and precisely as possible)**: See above

**Environment**:
- ksonnet version (use `ks version`):

```
$ ks version
ksonnet version: 0.10.0-alpha.1
jsonnet version: v0.10.0
client-go version: 1.8
```

bryanl:
* Create separate components for all items in YAML manifest
* Remove component index because it isn't needed

Fixes #437


actions/import.go
```diff
@@ -194,6 +194,81 @@ func (i *Import) importFile(fileName string) error {
...
+		if err != nil {
+			return err
+		}
+
+		name, ok := val.(string)
+		if !ok {
+			return errors.Errorf("unable to find metadata name of object in %s", fileName)
+		}
+
+		componentName := fmt.Sprintf("%s-%s", strings.ToLower(ts.Kind()), name)
```
redbaron: unfortunately, Kind + Name is not unique identifier :( Look at Voyager Ingress for example: https://github.com/appscode/voyager/blob/master/docs/examples/ingress/types/loadbalancer/ing.yaml

it clashes with builtin Ingress.

Kind + Name + Apiversion would be unique, but at least builtin  resources (Deployment, Job, etc) somehow automagically coalesce into same resource, so for them Kind + Name  is unique combination.

bryanl: This is a good catch, I took a look at the built in resources and there wasn't an issue. I will follow up with a change that will handle custom resources as well.

bryanl: It's now kind-name-random.

Adding apiVersion could have made it unbearable. If this doesn't work out, We'll revisit this.

PR #497, opened 2018-04-27
Allow forcing a parameter value to string

### LINKED ISSUE #492: ks param set converts values consisting of string of digits to numerics
/kind bug

**What happened**:
Trying to create parameter with a string of digits (e.g. "0", "00", "1", "2"), but ksonnet removes all double and single quotes from the all digit strings in the param file.

It should be possible to create parameter with values like "0", "00" and have them stay as literal strings instead of being converted into numerics.

**How to reproduce it (as minimally and precisely as possible)**:

`ks param set component_name x "0"`

view the params file and see the double quotes removed:
    x: 0,

Using the same approach it is possible to completely break "ks param list" like this:

`ks param set component_name x "00"`

Try to list params

`ks param list component_name`

Get error: ERROR get all component params for : 20:11-12 Expected a comma before next field.

Manually remove the x: 00 value from params.libsonnet file to restore "ks param" functionality.

**Anything else we need to know?**:

It does not seem possible to use ", \", ', \', "", '', etc. to force ks param set to keep the quotes. 

For example, using 

`ks param set component_name x \'0\'
`
The value added to the params file contains multiple quotes:

x: "'0'",

**Environment**:
- ksonnet version (use `ks version`):

ksonnet version: 0.9.2
jsonnet version: v0.9.5
client-go version: 1.8

- Kubernetes version (use `kubectl version`):

Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.1", GitCommit:"d4ab47518836c750f9949b9e0d387f20fb92260b", GitTreeState:"clean", BuildDate:"2018-04-12T14:26:04Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.6", GitCommit:"9f8ebd171479bec0ada837d7ee641dec2f8c6dd1", GitTreeState:"clean", BuildDate:"2018-03-21T15:13:31Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}

### LINKED ISSUE #496: Allow type hints for setting parameter values
In some cases the preferred type for a value can't be inferred through the normal means. In these cases, allow the user to supply a type hint to say what the value is.

e.g. 

`ks param set component key 1` 

In this case, `key` would be set to 1. This might always not be desired as the user might want `key` to be `"1"`. To support the user, introduce a new flag, `--as-string` to force the value to be a string.

#492

bryanl:
Adds ability to force a paramter value to string using the `--as-string`
flag.

e.g.

```
[I] mule · value-string-type-hint± ⟩ ks param set comp aDigit 1
[I] mule · value-string-type-hint± ⟩ ks param list comp
COMPONENT PARAM    VALUE
========= =====    =====
comp      aDigit   1
comp      bindings [{"name":"name","role":"role","users":["user1","user2"]}]
[I] mule · value-string-type-hint± ⟩ ks param set comp aDigit 1 --as-string
[I] mule · value-string-type-hint± ⟩ ks param list comp
COMPONENT PARAM    VALUE
========= =====    =====
comp      aDigit   "1"
comp      bindings [{"name":"name","role":"role","users":["user1","user2"]}]
```

Fixes #496
Fixes #492

### DIFF
--- a/pkg/actions/actions.go
+++ b/pkg/actions/actions.go
@@ -29,6 +29,8 @@ const (
 	OptionApp = "app"
 	// OptionArguments is arguments option. Used for passing arguments to prototypes.
 	OptionArguments = "arguments"
+	// OptionAsString is asString. Used for setting values as strings.
+	OptionAsString = "as-string"
 	// OptionClientConfig is clientConfig option.
 	OptionClientConfig = "client-config"
 	// OptionComponentName is a componentName option.
--- a/pkg/actions/param_set.go
+++ b/pkg/actions/param_set.go
@@ -44,6 +44,7 @@ type ParamSet struct {
 	rawValue string
 	global   bool
 	envName  string
+	asString bool
 
 	getModuleFn    getModuleFn
 	resolvePathFn  func(a app.App, path string) (component.Module, component.Component, error)
@@ -62,6 +63,7 @@ func NewParamSet(m map[string]interface{}) (*ParamSet, error) {
 		rawValue: ol.LoadString(OptionValue),
 		global:   ol.LoadOptionalBool(OptionGlobal),
 		envName:  ol.LoadOptionalString(OptionEnvName),
+		asString: ol.LoadOptionalBool(OptionAsString),
 
 		getModuleFn:    component.GetModule,
 		resolvePathFn:  component.ResolvePath,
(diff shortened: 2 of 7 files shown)


shaoxt:
It doesn't work in ks version 0.13.1

params.libsonnet:

local params = import '../../components/params.libsonnet';

params + {
  components+: {
    sample+: {
      env: 'qal',
      assetId: 6198096579474420594,
    },
  },
}

When I run, it still use integer,

/usr/local/bin/ks param set sample assetId 6198096579474420594 --as-string --env qal

ks version
ksonnet version: 0.13.1
jsonnet version: v0.11.2
client-go version: kubernetes-1.10.4

rahul8590:
It doesn't work for boolean values too. 
I tried the following 
`ks param set sample_var "true" --as-string -env <evn>`
expecting the value to be saved as string, but sets as `bool` type. 

ks version
ksonnet version: 0.13.1

PR #501, opened 2018-05-02
Add component paths to jpath

bryanl:


redbaron:
components can use relative imports to reach their own files. cross components imports can't work reliably (import conflicts, version changes etc)

bryanl:
@redbaron The reason this PR is WIP is because I'm not sure this is a good idea. With the introduction of modules, there is a conflict with projects that have decided to use directories under components.

PR #560, opened 2018-05-24
Add apiVersion/kind to app.override.yaml

### LINKED ISSUE #513: app.override.yaml needs kind and version
Update `app.override.yaml` to include a kind and version to let `ks` know exactly what it is dealing with.

bryanl:
Adds identifying information to app override configuration

example:

```
apiVersion: 0.1.0
kind: ksonnet.io/app-override
registries:
  container:
    protocol: fs
    uri: ../../../../../../Development/heptio/ksonnet-container
```

Fixes #513

### DIFF
--- a/pkg/app/override.go
+++ b/pkg/app/override.go
@@ -15,8 +15,63 @@
 
 package app
 
+import (
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+)
+
+const (
+	// overrideKind is the override resource type.
+	overrideKind = "ksonnet.io/app-override"
+	// overrideVersion is the version of the override resource.
+	overrideVersion = "0.1.0"
+)
+
 // Override defines overrides to ksonnet project configurations.
 type Override struct {
+	Kind         string           `json:"kind"`
+	APIVersion   string           `json:"apiVersion"`
 	Environments EnvironmentSpecs `json:"environments,omitempty"`
 	Registries   RegistryRefSpecs `json:"registries,omitempty"`
+}
+
+// Validate validates an Override.
+func (o *Override) Validate() error {
+	if o.Kind != overrideKind {
+		return errors.Errorf("app override has unexpected kind")
+	}
+
+	if o.APIVersion != overrideVersion {
+		return errors.Errorf("app override has unexpected apiVersion")
+	}
+
+	return nil
+}
+
+// IsDefined returns true if the override has environments or registries defined.
+func (o *Override) IsDefined() bool {
+	return len(o.Environments) > 0 || len(o.Registries) > 0
+}
+
+// SaveOverride saves the override to the filesystem.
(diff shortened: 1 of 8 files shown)


redbaron:
IMHO override shouldn't apply if it overrides wrong apiVersion.

bryanl:
@redbaron You are right. The code now checks apiVersion and kind.

pkg/app/override.go
```diff
@@ -15,8 +15,63 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+)
+
+const (
+	// overrideKind is the override resource type.
+	overrideKind = "ksonnet.io/app-override"
+	// overrideVersion is the version of the override resource.
+	overrideVersion = "0.1.0"
```
shomron: Should we reuse `DefaultAPIVersion` from `schema.go`? Alternatively, should we be comparing to the loaded API version of the current `App` spec, rather than a separate hardcoded value?

pkg/app/override.go
```diff
@@ -15,8 +15,63 @@
...
+// SaveOverride saves the override to the filesystem.
+func SaveOverride(encoder Encoder, fs afero.Fs, root string, o *Override) error {
+	if o == nil {
+		return errors.New("override was nil")
+	}
+
+	o.APIVersion = overrideVersion
+	o.Kind = overrideKind
+
+	f, err := fs.OpenFile(overridePath(root), os.O_RDWR|os.O_CREATE, DefaultFilePermissions)
```
shomron: Is `os.O_WRONLY|os.O_CREATE` sufficient?

pkg/app/override.go
bryanl: No, DefaultAPIVersion is the version for `app.yaml`. These files are of different types and can be versioned separately.

pkg/app/override.go
```diff
@@ -15,8 +15,63 @@
...
 	Registries   RegistryRefSpecs `json:"registries,omitempty"`
 }
+
+// Validate validates an Override.
+func (o *Override) Validate() error {
+	if o.Kind != overrideKind {
+		return errors.Errorf("app override has unexpected kind")
+	}
+
+	if o.APIVersion != overrideVersion {
```
shomron: We can plumb this for multiple versions when there are multiple versions 😄

pkg/app/schema.go
```diff
@@ -171,25 +169,18 @@ func write(fs afero.Fs, appRoot string, spec *Spec) error {
 		return errors.Wrap(err, "write app.yaml")
 	}
 
-	if err = cleanOverride(fs, appRoot); err != nil {
+	if err = removeOverride(fs, appRoot); err != nil {
```
shomron: I have questions around transactionality - if `SaveOverride` fails below, is it ok that we overwrote the `app.yaml` above?

bryanl: That's an issue which should be tracked. I'm creating an issue for it.

Issue #273, opened 2018-01-07
Is it possible to specify empty string as default for parameter value?


jlewi:
Is there a way to specify that the default value for a string parameter in a prototype should be the empty string?

bryanl:
If I'm understanding your question correctly, you should be able to do:

```sh
ks generate prototype name --paramName=""
```

a more concrete example with our redis prototype:

```sh
ks generate redis-stateless redis --redisPassword=""  
```
And this produces in params.libsonnet

```javascript
{
  redis: {
    name: "redis",
    redisPassword: "",
  },
}
```

Or did I misunderstand your question?

jlewi:
I'm creating a new part with a new protoype. So I have a jsonnet file like so

```
// @apiVersion 0.1
// @name io.ksonnet.pkg.iap
// @description Provides K8s resources needed to use IAP with JupyterHub on GKE.
// @shortDescription IAP resources for JupyterHub on GKE.
// @param name string Name for the component
// @optionalParam namespace string default Namespace
// @optionalParam certsSecretName string null the name of the secret containing the SSL certificates.
```
I'd like the default value for `certsSecretName` to be the empty string. But as far as I can tell that's not possible because it will end up using the first word of my description as the default. 

As a work around I set the default value to null and then check if the string is "null" and then treat it as the empty string.

bryanl:
That's what some of the existing prototypes do:

https://github.com/ksonnet/parts/blob/master/incubator/redis/prototypes/redis-stateless.jsonnet#L20

```js
local optionalSecret =
  if redisPassword != "null"
  then redis.parts.secret(name, redisPassword)
  else null;
```

This pattern may be ripe for disruption, but that does work now.

hausdorff:
@jlewi You cannot, we're not yet escaping quotes as part of the default values. I thought we'd recorded this as a part of #69, but I guess not. We should fix it.

hausdorff:
Probably not slated for 0.9.0, though. THoughts @jessicayuen @bryanl ?

jessicayuen:
We should fix this for 0.9.0.

bryanl:
Let's pull up and reprioritize tomorrow.

PR #573, opened 2018-05-29
wip: Use values file to supply values for prototypes

### LINKED ISSUE #550: Create values flag to supply values to `ks prototype use`
Values for prototypes can be complex. To allow for more flexibility, all user to use a value file to supply the options.

Usage: `ks prototype use <id> --values values-file`

Where `values-file` is file with keys and values:

```
key1=value1
key2=value2
```

bryanl:
```sh
ks generate deployed-service deployed-service5 --values-file /tmp/values-file
```

Fixes #550


pkg/clicmd/prototype_use.go
```diff
@@ -75,6 +75,16 @@ expand prototypes into Jsonnet files.
 command line flags, such as ` + " `--image` " + `in the example above. Note that
 different prototypes support their own unique flags.
 
+You can also use a values file to supply values to the prototype. The format for the
+file is a key and value per line.
+
+		key1=value1
```
shomron: We should be explicit about what characters are and aren't allowed in keys and values. For example, do we need to support `=` characters in values?

pkg/prototype/values.go
```diff
@@ -0,0 +1,226 @@
...
+
+		return vff.Extract()
+	}
+
+	vff := newValuesFromFlags(proto, flags)
+	return vff.Extract()
+}
+
+type valuesFromFile struct {
+	fs    afero.Fs
```
shomron: Why is this needed? Is the `io.Reader` abstraction enough?

pkg/prototype/values_test.go
```diff
@@ -0,0 +1,127 @@
...
+				"key2": `"default"`,
+			},
+		},
+		{
+			name:  "values: invalid",
+			p:     proto,
+			flags: BindFlags(proto),
+			args: []string{
+				"--name", "name",
+				"--values-file", "/files/invalid",
```
shomron: If `=` is prohibited in values, can we add a test case capturing that?

bryanl: It's not prohibited, but the string is split into two parts. I'll add a test case for this.

pkg/clicmd/prototype_use.go
bryanl: ksonnet accept any unicode character if the Param is designated as a string.

pkg/prototype/values_test.go
shomron: It looks like we're explicitly checking for a two-way split, which would preclude additional `=`:

```
		parts := strings.SplitN(s, "=", 2)
		if len(parts) != 2 {
			return nil, errors.Errorf("%q is invalid. value format is key=value", s)
		}
```

bryanl: No, that's not how it would work. Here's an example:

https://play.golang.org/p/_uR5RufO667

shomron: Ah, so that is the difference between `Split` and `SplitN` 🙂
I misunderstood.

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+
+		quoted, err := param.Quote(val)
+		if err != nil {
+			return nil, err
+		}
+		values[param.Name] = quoted
+	}
+
+	if len(missingRequired) > 0 {
+		return nil, errors.Errorf("failed to instantiate prototype %q. The following required parameters are missing:\n%s", proto.Name, missingRequired.PrettyString(""))
```
shomron: `proto.Name` -> `vff.proto.Name`

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+
+import (
+	"bufio"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+	"github.com/spf13/pflag"
+	"google.golang.org/grpc/encoding/proto"
```
shomron: I believe this isn't needed.

Gopkg.lock
```diff
@@ -773,6 +773,15 @@
   revision = "150dc57a1b433e64154302bdc40b6bb8aefa313a"
   version = "v1.0.0"
 
+[[projects]]
+  name = "google.golang.org/grpc"
```
shomron: Was this brought in due to an over-eager `goimports`

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil, err
+	}
+
+	name, err := flags.GetString("name")
+	if err != nil {
+		return nil, errors.New("name for prototype is required")
+	}
+
+	if valuesFile != "" {
+		vff, err := newValuesFromFile(fs, p, name, valuesFile)
```
shomron: We should document somewhere that valuesFile will take precedence over bare flags.

pkg/prototype/prototype.go
```diff
@@ -189,6 +200,25 @@ type ParamSchema struct {
...
+	var v string
+	if ps.Default != nil {
+		v = *ps.Default
+	}
+
+	if v == "" {
+		v = value
+	}
+
+	return ps.Quote(v)
```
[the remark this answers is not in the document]
bryanl: I think this should be filed as a bug and handled separately.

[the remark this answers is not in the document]
shomron: 👍

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+	values := map[string]string{}
+	missingRequired := ParamSchemas{}
+
+	for _, param := range vff.p.Params {
+		val, err := vff.flags.GetString(param.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		if param.Default != nil {
```
shomron: `param.IsRequired()`

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+
+		if parts[1] == "" {
+			return nil, errors.Errorf("%q is invalid. value format is key=value", s)
+		}
+
+		key := strings.TrimSpace(parts[0])
+		value := strings.TrimSpace(parts[1])
+
+		m[key] = value
```
shomron: Duplicate keys are not detected.

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+
+	if len(missingRequired) > 0 {
+		return nil, errors.Errorf("failed to instantiate prototype %q. The following required parameters are missing:\n%s", vff.p.Name, missingRequired.PrettyString(""))
+	}
+
+	return values, nil
+}
+
+func BindFlags(p *Prototype) *pflag.FlagSet {
+	fs := pflag.NewFlagSet("preview", pflag.ContinueOnError)
```
shomron: What does `preview` signify? Flags for the `prototype preview` command?

pkg/prototype/values_test.go
```diff
@@ -0,0 +1,141 @@
...
+			},
+			{
+				Name:    "key2",
+				Type:    String,
+				Default: ksstrings.Ptr("default"),
+			},
+		},
+	}
+
+	cases := []struct {
```
shomron: We could use one more test to show precedence logic between bare flags and values file.

pkg/prototype/prototype.go
```diff
@@ -189,6 +200,25 @@ type ParamSchema struct {
 	Type        ParamType `json:"type"`
 }
 
+// IsRequired returns if this parameter is required. Parameters are required
+// if the default value is nil.
+func (ps *ParamSchema) IsRequired() bool {
+	return ps.Default == nil
+}
+
+func (ps *ParamSchema) QuotedValue(value string) (string, error) {
```
shomron: I propose renaming this method `QuotedWithDefault` - as is the difference vs. `Quote` is not clear to the reader.

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+				missingRequired = append(missingRequired, param)
+				continue
+			}
+		}
+
+		if _, ok := values[param.Name]; ok {
+			return nil, errors.Errorf("prototype %q has multiple parameters with name %q", vff.p.Name, param.Name)
+		}
+
+		quoted, err := param.Quote(val)
```
shomron: Should probably call `QuotedValue` / `QuotedWithDefault`.

pkg/prototype/prototype.go
shomron: Is it ever valid to set a parameter to an empty string? If so, we'll need a way to distinguish between that case and the case where no value was given and default should be used.

pkg/prototype/values.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	return &valuesFromFile{
+		p:    p,
+		name: name,
+		r:    f,
+	}, nil
+}
+
+func (vff *valuesFromFile) Extract() (map[string]string, error) {
```
shomron: I think we should also error out on unexpected keys in the values file, where unexpected means no match in the schema for this prototype. The current state lets the default be used silently even though the user intended to override.

pkg/prototype/prototype.go
shomron: related: https://github.com/ksonnet/ksonnet/issues/273

Issue #522, opened 2018-05-08
Introduce helm component


bryanl:
ksonnet should be able to consume helm charts as a component type.

## [Helm Registry](https://docs.google.com/document/d/1FcK8pDqj_9EFjNVr9e_t9WemouQpS9ix-r0AxNDB_nM/edit?usp=sharing)

- [x] ksonnet will expose Helm repositories as registries
- [x] Helm based registries can expose their contents as prototypes #664
- [x] Helm registries can manage multiple Chart releases simultaneously. #672
- [x] Helm registry should attempt to cache entries

### Helm upgrade

A ksonnet user should be able to upgrade a chart to the specific version or the latest release.

burdiyan:
One thing to take into account that is specific to Helm is hooks: https://github.com/kubernetes/helm/blob/master/docs/charts_hooks.md

I think that replicating this functionality should not be a goal, so maybe Ksonnet should interact with Tiller to install Helm charts, instead of rendering them locally.

bryanl:
I'm actually leaning towards making helm charts registries and using the helm rendering engine to generate something ksonnet can understand. The goal for this is not to replicate all of helm inside of ksonnet, but to instead make charts (and their dependencies) available to ksonnet.

burdiyan:
Some of Charts inside stable Helm repository are using hooks as a crucial part of the chart. For example `mongodb-replicaset` chart is using post-install hook to form the actual ReplicaSet after all nodes are up.

bryanl:
@burdiyan after reviewing the hooks, the first version of this feature will not include support for hooks. ksonnet's operating model is declarative and hooks are imperative, so this fit is not obvious.

burdiyan:
@bryanl Make sense for the initial implementation. Although would be great if ksonnet fail when trying to import a chart with hooks. Or at least a warning.

bryanl:
@burdiyan that is the plan. Working out when  and where to fail right now.

PR #583, opened 2018-06-03
Create helm registry for ksonnet

bryanl:
Creates a helm registry that can treat helm charts as ksonnet parts.

[design doc](https://docs.google.com/document/d/1FcK8pDqj_9EFjNVr9e_t9WemouQpS9ix-r0AxNDB_nM/edit?usp=sharing)

#522

This change does the following:

If you have a helm registry defined in `app.yaml`, It will show up in the registry listing:

```
NAME        OVERRIDE PROTOCOL URI
====        ======== ======== ===
helm-stable          helm     https://kubernetes-charts.storage.googleapis.com
```

You can run `ks pkg list` and see charts in your helm repository

```
REGISTRY    NAME                           INSTALLED
========    ====                           =========
helm-stable acs-engine-autoscaler
helm-stable aerospike
helm-stable anchore-engine
helm-stable artifactory
helm-stable artifactory-ha
helm-stable aws-cluster-autoscaler
helm-stable bitcoind
helm-stable buildkite
helm-stable burrow
```

You can run `ks pkg install` to install a package:

```
$ ks pkg install helm-stable/acs-engine-autoscaler
INFO Retrieved 8 files
```

The pkg will then be vendored for you:

```
vendor/helm-stable/
└── acs-engine-autoscaler
    └── helm
        └── 2.2.0
            └── acs-engine-autoscaler
                ├── Chart.yaml
                ├── README.md
                ├── templates
                │   ├── NOTES.txt
                │   ├── _helpers.tpl
                │   ├── deployment.yaml
                │   └── secrets.yaml
                └── values.yaml

5 directories, 7 files
```


rimusz:
👍

rimusz:
would you mind to give comment access to the design doc, please?

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+			Path:    name,
+			Version: chart.Version,
+		}
+	}
+
+	return spec, nil
+}
+
+// MakeRegistryRefSpec returns app registry ref spec.
+func (h *Helm) MakeRegistryRefSpec() *app.RegistryRefSpec {
```
underrun: why is this called MakeRegistryRefSpec instead of RegistryRefSpec?

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	"github.com/pkg/errors"
+)
+
+// Helm is a Helm repository.
+type Helm struct {
+	app              app.App
+	spec             *app.RegistryRefSpec
+	repositoryClient HelmRepositoryClient
+	unarchiver       archive.Unarchiver
+	httpGet          httpGet
```
shomron: Why is this not encapsulated in the `repositoryClient`?

bryanl: Because I missed it. It should be in there.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	if rc == nil {
+		return nil, errors.New("helm repository client is nil")
+	}
+
+	h := &Helm{
+		app:              a,
+		spec:             registryRef,
+		repositoryClient: rc,
+		unarchiver:       ua,
+		httpGet:          http.Get,
```
shomron: We should support request timeouts - `http.DefaultClient` does not.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+
+// HelmRepositoryChart is metadata describing a Helm Chart in a repository.
+type HelmRepositoryChart struct {
+	Description string   `json:"description,omitempty"`
+	Name        string   `json:"name,omitempty"`
+	URLs        []string `json:"urls,omitempty"`
+	Version     string   `json:"version,omitempty"`
+}
+
+type httpGet func(string) (*http.Response, error)
```
shomron: Is this style preferred over a `getter` interface?

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+
+	"github.com/blang/semver"
+	"github.com/ghodss/yaml"
+	"github.com/ksonnet/ksonnet/pkg/app"
+	"github.com/ksonnet/ksonnet/pkg/parts"
+	"github.com/ksonnet/ksonnet/pkg/util/archive"
+	"github.com/pkg/errors"
+)
+
+// Helm is a Helm repository.
```
shomron: Is there a good helm repository spec we could link to from somewhere? The most comprehensive I found was [this](https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md)

bryanl: The might be the best there is.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func newHelmRepositoryClient(u string, hg httpGet) (*helmRepositoryClient, error) {
+	u, err := normalizeHelmURI(u)
+	if err != nil {
+		return nil, errors.Wrap(err, "normalizing Helm repository URL")
+	}
+
+	if hg == nil {
+		hg = http.Get
```
shomron: Same for timeouts.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	return &helmRepositoryClient{
+		url:     u,
+		httpGet: hg,
+	}, nil
+}
+
+func (hrc *helmRepositoryClient) Repository() (*HelmRepository, error) {
```
shomron: Let's document exported methods.

bryanl: This method isn't exported, but it should have documentation.

shomron: Fair enough, and agreed 😃

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	defer func() {
+		cerr := resp.Body.Close()
+		if cerr != nil {
+			err = cerr
+		}
+	}()
+
+	b, err := ioutil.ReadAll(resp.Body)
```
shomron: How do we feel about reading arbitrarily long bodies into memory?

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("repository returned unexpected response %s", resp.Status)
+	}
+
+	defer func() {
+		cerr := resp.Body.Close()
+		if cerr != nil {
+			err = cerr
```
shomron: What's the intention here? What is generally done with errors within `defer` blocks?

bryanl: This is an extremely hard error to trigger. If it does happen, return it instead of the initial error since it is most likely the cause of the problem. The reason it's a deferred function is because `defer resp.Body.Close()` will swallow the error.

shomron: But to pull that off correctly, would `error` need to be a named return value?

shomron: Also, if that's the plan it might be better to wrap any existing error rather than steam-rolling it.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+
+	if err := yaml.Unmarshal(b, &hre); err != nil {
+		return nil, errors.Wrap(err, "unmarshalling repository index.yaml")
+	}
+
+	return &hre, nil
+}
+
+func (hrc *helmRepositoryClient) Chart(name, version string) (*HelmRepositoryChart, error) {
+	repo, err := hrc.Repository()
```
shomron: This should probably be cached - do we want to download and parse multiple megabytes for each call?

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "retrieving repository")
+	}
+
+	if version == "" {
+		for _, chart := range repo.Latest() {
```
shomron: This seems like a lot of extra work - we're sorting all versions of all charts, then doing a linear search for the desired chart name. Why instead not sort just `repo.Charts[name]`?

pkg/registry/manager.go
```diff
@@ -60,6 +60,12 @@ func Locate(a app.App, spec *app.RegistryRefSpec) (Registry, error) {
 		return githubFactory(a, spec)
 	case ProtocolFilesystem:
 		return NewFs(a, spec)
+	case ProtocolHelm:
```
shomron: Is `registry add` support in scope for this PR?

bryanl: Nope. Don't want to do that until we have a complete solution.

pkg/registry/manager.go
```diff
@@ -60,6 +60,12 @@ func Locate(a app.App, spec *app.RegistryRefSpec) (Registry, error) {
...
+		client, err := newHelmRepositoryClient("https://kubernetes-charts.storage.googleapis.com", nil)
```
shomron: Should this be configurable? Or at least `const`? Or will that be part of the `registry add` work?

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	for _, u := range chart.URLs {
+		resp, err := h.httpGet(u)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			return nil, nil, errors.Errorf("unexpected HTTP status code %q when retrieving helm chart", resp.Status)
```
shomron: Some more context in this error (chart? url?) would be useful - I'm not clear on whether those are added up the stack.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+
+		defer func() {
+			cerr := resp.Body.Close()
+			if cerr != nil {
+				err = cerr
+			}
+		}()
+
+		handler := func(f *archive.File) error {
```
shomron: A comment along the lines of "file contents are returned to the caller via `onFile` handler would be useful to the uninitiated.

pkg/registry/manager.go
bryanl: It should be, but this is dependent on `registry add`. So for now, it is hard coded.

pkg/registry/helm.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+
+		if err = h.unarchiver.Unarchive(resp.Body, handler); err != nil {
+			return nil, nil, err
+		}
+	}
+
+	refSpec := &app.LibraryRefSpec{
+		Name:     partAlias,
+		Registry: h.Name(),
```
shomron: Conspicuously missing is `GitVersion` - should we discuss how we might better abstract away the version details while allowing them to be persisted in `app.yaml`?

bryanl: There is not GitVersion if there is not GitRepo. This is an implementation detailed that leaked before registries were understood.

pkg/registry/helm.go
bryanl: I agree, going to add another issue for this.

Makefile
```diff
@@ -16,7 +16,7 @@
 VERSION?=dev-$(shell date +%FT%T%z)
 KS_BIN?=ks
 
-APIMACHINERY_VER := $(shell dep status | grep k8s.io/apimachinery | awk '{print $$3}')
```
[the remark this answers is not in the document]
underrun: Another option here if we we want to keep the flexibility without the terrible build time: 
`APIMACHINERY_VER := $(shell grep -B1 k8s.io/apimachinery Gopkg.lock | head -n1 | cut -d'"' -f2)`

bryanl:
@ksonnet/reviewers  is this ready to go? Any more changes?

pkg/helm/client.go
```diff
@@ -0,0 +1,247 @@
...
+	// Fetch fetches URIs for a chart.
+	Fetch(uri string) (io.ReadCloser, error)
+}
+
+// Repository is metadata describing the contents of a Helm repository.
+type Repository struct {
+	Charts map[string][]RepositoryChart `json:"entries,omitempty"`
+}
+
+// FooLatest returns the latest version of charts in a Helm repository.
```
shomron: Minus Foo

pkg/helm/client.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	return v1.Compare(v2) == 1
+}
+
+type Getter interface {
+	Get(string) (*http.Response, error)
+}
+
+type getter struct {
```
shomron: I propose renaming to `httpGetter` - it will be clearer to read vs. the abstract getter interface.

pkg/helm/client.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return &HTTPClient{
+		url:    u,
+		getter: hg,
+	}, nil
+}
+
+// Repository returns the Helm repository's content.
+func (hrc *HTTPClient) Repository() (*Repository, error) {
+	resp, err := hrc.getter.Get(hrc.url.String())
```
shomron: Should we use `Fetch()` from below here?

pkg/helm/client.go
```diff
@@ -0,0 +1,247 @@
...
+		if version == chart.Version {
+			return &chart, nil
+		}
+	}
+
+	return nil, errors.Errorf("chart %q with version %q was not found", name, version)
+}
+
+// Fetch fetches URLs from a repository. If uri is a path, it will use the client URL as the base.
+func (hrc *HTTPClient) Fetch(uri string) (io.ReadCloser, error) {
```
shomron: Should this be exported?

shomron:
Added a last round of comments. None need to be hard blockers.

pkg/helm/client.go
bryanl: Yes, because the Helm registry uses it to resolve files and it is in another package.

PR #588, opened 2018-06-04
add docker image build support

### LINKED ISSUE #291: Implement standard dockerfile
/kind feature

In the past, we've released docker images _ad hoc_. People ask us for this pretty frequently, so we should (1) implement a standard dockerfile, and (2) document it as part of the release checklist.

@chuckha has been asking for it internally also.

underrun:
closes #291 

rough edges around running with mounts and permissions, but bones are here. could use advice on creating an alias for this - had some trouble because it needs pwd. thoughts?


Makefile
```diff
@@ -45,6 +46,9 @@ ks:
 docs:
 	$(DOC_GEN_FILE)
 
+docker-image:
+	docker build -t ks --build-arg LD_FLAGS="$(LD_FLAGS)" .
```
shomron: We should probably tag according to the ksonnet version as well. Maybe `git describe --tags` with or without the `v` prefix?

docs/docker-image-build.md
```diff
@@ -0,0 +1,20 @@
...
+dep ensure
+make docker-image
+```
+
+# Running `ks` via docker
+
+In order to run via docker, the `ks` process needs access to a kubernetes config file, certificates, and the current working directory. We set these up using bind mounts:
+
+```bash
+docker run -e KUBECONFIG="${KUBECONFIG}" --mount type=bind,source="${KUBECONFIG}",target="${KUBECONFIG}" --mount type=bind,source="${HOME}/.minikube",target="${HOME}/.minikube" --mount type=bind,source="$(pwd)",target="$(pwd)" -w "$(pwd)" ks --help
```
shomron: This fails for people like me who lack a `KUBECONFIG` variable.
Let's supply a default:

`${KUBECONFIG}` -> `${KUBECONFIG:-${HOME}/.kube/config}`

shomron: The minikube dependency is not always going to be correct - perhaps what we really give is a bash function that parses out the certificate paths from `${KUBECONFIG}` and builds the arguments that way? Or is that overkill?

Makefile
underrun: if you look in the LD_FLAGS it grabs `git rev-parse HEAD` which gives you sha1 ... but do you mean for human consumption like having a ks images  named ks-0.11.0 or something along those lines?

docs/docker-image-build.md
underrun: oh man i know - i actually had the exact line you suggest for `KUBECONFIG` with default ... and i had a CERTDIR with default to ${HOME}/.kube ... and i talked to @bryanl about parsing the cert file but apparently there are versions of the cert specification that are links on which actions need to be taken so ... that's exciting.

the ergonomics of this are why i didn't include something to add to `.bash_aliases` - it's way to user specific to do something simple. we can make this work intuitively but it might have diminishing returns.

Makefile
shomron: Yeah look how images are tagged on DockerHub (where these images should ultimately be pushed via Jenkins). I'd only include sha for non-release tags.

docs/docker-image-build.md
shomron: Agreed. Maybe we attack this with documentation, describing what needs to be mounted logically and giving an example.

underrun: lol - yeah - this is the docs that i meant that to be - i'll add a couple other examples and note failure cases as well to help people navigate.

Makefile
```diff
@@ -16,12 +16,14 @@
 VERSION?=dev-$(shell date +%FT%T%z)
 KS_BIN?=ks
 
-APIMACHINERY_VER := $(shell dep status | grep k8s.io/apimachinery | awk '{print $$3}')
+APIMACHINERY_VER := $(shell grep -B1 k8s.io/apimachinery Gopkg.lock | head -n1 | cut -d'"' -f2)
 REVISION=$(shell git rev-parse HEAD)
+GIT_TAG=$(shell git describe --tags)
```
bryanl: Would it make sense to only set `GIT_TAG` if is needed?

Makefile
```diff
@@ -39,12 +41,20 @@ INTEGRATION_TEST_FIXTURES = ./fixtures
...
+	touch Gopkg.lock
+	$(eval APIMACHINERY_VER := $(shell grep -B1 k8s.io/apimachinery Gopkg.lock | head -n1 | cut -d'"' -f2))
+
+ks: Gopkg.lock
 	$(GO) build -o $(KS_BIN) $(GO_FLAGS) ./cmd/ks
 
 docs:
 	$(DOC_GEN_FILE)
 
+docker-image: Gopkg.lock
```
bryanl: This should be be added to `.PHONY` at the bottom of the file? (and also `install` should be as well)

docs/docker-image-build.md
```diff
@@ -0,0 +1,29 @@
+# Build a `ks` docker image
```
bryanl: @GuessWhoSamFoo What do you think about this text?

docs/docker-image-build.md
```diff
@@ -0,0 +1,29 @@
...
+
+In order to run via docker, the `ks` process needs access to a kubernetes config file, certificates, and the current working directory. We can set these up using mounts. Here's what it looks like running on a local cluster with minikube:
+
+```bash
+docker run -e KUBECONFIG="~/.kube/config" --mount type=bind,source="~/.kube/conifg",target="~/.kube/config" --mount type=bind,source="~/.minikube",target="~/.minikube" --mount type=bind,source="$(pwd)",target="$(pwd)" -w "$(pwd)" ks --help
+```
+
+This sets the $KUBECONFIG environment variable inside the container, mounts the config and the directory holding certificates (which can be found inside the kubeconfig), and mounts the current working directory so that the `ks` binary knows where to work.
+
+It is also possible to use as an alias:
```
GuessWhoSamFoo: Clarify the intention of setting an alias. e.g:

Optionally, set an alias to shorten the verbose command if using the docker image locally.

docs/docker-image-build.md
```diff
@@ -0,0 +1,29 @@
...
+go get github.com/ksonnet/ksonnet
+
+# Build and install binary under shortname `ks` into $GOPATH/bin
+cd "${GOPATH}/src/github.com/ksonnet/ksonnet"
+make docker-image
+```
+
+# Running `ks` via docker
+
+In order to run via docker, the `ks` process needs access to a kubernetes config file, certificates, and the current working directory. We can set these up using mounts. Here's what it looks like running on a local cluster with minikube:
```
GuessWhoSamFoo: Instead of "We can set these up using mounts"

Pass the `--mount` flag to capture the path of these files.

docs/docker-image-build.md
GuessWhoSamFoo: Nit: Referencing Docker outside the context of a CLI should be upper case.

Makefile
underrun: so in Makefiles variables set with = are lazily evaluated ... so it really doesn't hurt anything to make it available here as it won't run `git describe` unless you are building a docker image. but i don't mind moving this into the target if that seems cleaner?

Makefile
underrun: will do

PR #592, opened 2018-06-07
adding component selection for diff

### LINKED ISSUE #581: `ks diff` doesn't support the `-c` option anymore
/kind bug

**What happened**:
`ks diff` used to support a `-c` option to work on a specific component only.
Starting with 0.11, ksonnet throws the following error
```
$ ks diff development -c 'aws-kube-dns'
ERROR unknown shorthand flag: 'c' in -c
$
```

ksonnet should diff only the specified component

**Environment**:
```
$ ks version
ksonnet version: 0.11.0
jsonnet version: v0.10.0
client-go version:
```

bryanl:
This change adds components for diffs. You can now run:

```
ks diff local:default remote:default -c <component name>
```

To diff a component in a module, specify the component with a module name. e.g component (for a module in the root component) or nested.component (for a module in nested/ under components)

In addition, module support throughout the ks was updated

* Modules allow ksonnet users to separate their configurations under components.
* Modules are dot separated paths that correlate to their location on the file system. e.g module a is in /app/components/a
* Modules can be deeply nested. e.g components in /app/components/nested/deeply would be in module nested.deeply
* Modules are now added as a label to kubernetes resources in `ksonnet.io/component`. This will allow the user to determine which component created a resource

Fixes #581

### DIFF
--- a/pkg/actions/diff.go
+++ b/pkg/actions/diff.go
@@ -111,7 +111,7 @@ func (d *Diff) Run() error {
 				return err
 			}
 		case strings.HasPrefix(t, "-"):
-			diffRemoveColor.Fprintln(&buf, t)
+			_, err = diffRemoveColor.Fprintln(&buf, t)
 			if err != nil {
 				return err
 			}
--- a/pkg/actions/diff.go
+++ b/pkg/actions/diff.go
@@ -54,8 +54,9 @@ type Diff struct {
 	clientConfig *client.Config
 	src1         string
 	src2         string
+	components   []string
 
-	diffFn func(app.App, *client.Config, *diff.Location, *diff.Location) (io.Reader, error)
+	diffFn func(app.App, *client.Config, []string, *diff.Location, *diff.Location) (io.Reader, error)
 
 	out io.Writer
 }
@@ -69,6 +70,7 @@ func NewDiff(m map[string]interface{}) (*Diff, error) {
 		clientConfig: ol.LoadClientConfig(),
 		src1:         ol.LoadString(OptionSrc1),
 		src2:         ol.LoadOptionalString(OptionSrc2),
+		components:   ol.LoadStringSlice(OptionComponentNames),
 
 		diffFn: diff.DefaultDiff,
 
@@ -91,7 +93,7 @@ func (d *Diff) Run() error {
 	}
 	location2 := diff.NewLocation(d.src2)
 
-	r, err := d.diffFn(d.app, d.clientConfig, location1, location2)
+	r, err := d.diffFn(d.app, d.clientConfig, d.components, location1, location2)
 	if err != nil {
 		return err
 	}
@@ -104,9 +106,15 @@ func (d *Diff) Run() error {
 
 		switch {
 		case strings.HasPrefix(t, "+"):
-			diffAddColor.Fprintln(&buf, t)
+			_, err = diffAddColor.Fprintln(&buf, t)
+			if err != nil {
+				return err
+			}
 		case strings.HasPrefix(t, "-"):
 			diffRemoveColor.Fprintln(&buf, t)
+			if err != nil {
+				return err
+			}
 		default:
 			fmt.Fprintln(&buf, t)
 		}
--- a/pkg/actions/import.go
+++ b/pkg/actions/import.go
@@ -54,7 +54,7 @@ type Import struct {
 	module string
 	path   string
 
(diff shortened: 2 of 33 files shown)


pkg/actions/diff.go
```diff
@@ -104,9 +106,15 @@ func (d *Diff) Run() error {
...
 		switch {
 		case strings.HasPrefix(t, "+"):
-			diffAddColor.Fprintln(&buf, t)
+			_, err = diffAddColor.Fprintln(&buf, t)
+			if err != nil {
+				return err
+			}
 		case strings.HasPrefix(t, "-"):
 			diffRemoveColor.Fprintln(&buf, t)
+			if err != nil {
```
underrun: line above this doesn't do `_, err =` like previous case - missed this?

pkg/actions/import.go
```diff
@@ -248,15 +248,15 @@ func (i *Import) createYAML(fileName, base, ext string) error {
 func (i *Import) createComponentFromData(name, data string, templateType prototype.TemplateType) error {
 	componentParams := params.Params{}
 
+	var moduleName string
 	switch i.module {
```
underrun: why keep this a switch when there's only one case? is a default-only switch a common golang idiom?

bryanl: 1. Because 0 values are useful in Go. It's the same thing as:

```go
var moduleName
if i.module != "" && i.module != "/" {
  moduleName = i.module
}
```
2. It's not a default only switch. There is a case statement in there too.

pkg/pipeline/pipeline.go
```diff
@@ -21,17 +21,18 @@ import (
 	"io"
 	"path/filepath"
 	"regexp"
-
-	"github.com/ksonnet/ksonnet/pkg/util/k8s"
-	"github.com/ksonnet/ksonnet/pkg/util/strings"
+	gostrings "strings"
```
shomron: I'm going to say stdlib should win the contest of _which package doesn't get aliased_.

pkg/actions/prototype_preview.go
```diff
@@ -121,6 +121,8 @@ func (pp *PrototypePreview) Run() error {
 func bindPrototypeParams(p *prototype.Prototype) *pflag.FlagSet {
 	fs := pflag.NewFlagSet("preview", pflag.ContinueOnError)
 
+	fs.String("module", "", "Component module")
```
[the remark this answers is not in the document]
bryanl: Is this possible?

pkg/pipeline/pipeline.go
bryanl: Why? I don't think there is a hard and fast rule.

pkg/component/jsonnet.go
```diff
@@ -68,11 +67,11 @@ func (j *Jsonnet) Name(wantsNameSpaced bool) string {
 		return name
 	}
 
-	if j.module == "/" {
+	if j.module == "/" || j.module == "" {
```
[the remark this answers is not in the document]
bryanl: No, it is `/`

pkg/cluster/object.go
```diff
@@ -62,8 +64,14 @@ func DefaultResourceInfo(namespace string, config clientcmd.ClientConfig) *resou
 		Flatten().
 		IncludeUninitialized(false).
 		RequireObject(true).
-		Latest().
-		Do()
+		Latest()
+
+	if len(components) > 0 {
+		selector := fmt.Sprintf("%s in (%s)", clustermetadata.LabelComponent, strings.Join(components, ","))
```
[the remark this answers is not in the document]
bryanl: Will this hold up this PR?

pkg/pipeline/pipeline.go
shomron: General familiarity and principal of least surprise. But this is a matter of preference, and only a suggestion :)

pkg/cluster/object.go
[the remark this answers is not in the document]
shomron: Doesn't have to. I can open an issue to harden this.

pkg/actions/prototype_preview.go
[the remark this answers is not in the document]
shomron: We could detect the collision and alert the user.

redbaron:
@bryanl , this change doesn't take into account that not all resources can have `.metadata.labels`, for instance `List` do not have it, opened https://github.com/ksonnet/ksonnet/issues/891

PR #600, opened 2018-06-08
Add `--installed` flag to pkg list

bryanl:
Adds a new flag to `pkg list` that only lists installed packages.

### DIFF
--- a/pkg/actions/pkg_list.go
+++ b/pkg/actions/pkg_list.go
@@ -43,7 +43,9 @@ func RunPkgList(m map[string]interface{}) error {
 
 // PkgList lists available registries
 type PkgList struct {
-	app            app.App
+	app           app.App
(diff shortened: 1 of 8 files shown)


pkg/actions/pkg_list.go
```diff
@@ -109,3 +112,12 @@ func (pl *PkgList) Run() error {
 	t.AppendBulk(rows)
 	return t.Render()
 }
+
+func (pl *PkgList) addRow(regName, libName string, isInstalled bool) []string {
+	row := []string{regName, libName}
+	if isInstalled {
+		row = append(row, pkgInstalled)
```
underrun: Do we really need the star if we are only listing installed things?

bryanl: I wanted to keep the column headings the same.

PR #608, opened 2018-06-12
Recover from panic when redifining prototype param

### LINKED ISSUE #607: Warn user when prototype parameters conflict with ksonnet parameters
When binding parameter flags from user-defined prototypes, it is possible for there to be a conflict between user parameter names and ksonnet-defined parameter names, e.g. `--module`.

These cases should be detected and the user should be warned about the conflict.

bryanl:
Prototypes now have builtin parameters (module)

If a prototype defined a parameter named module, recover
the panic that will occur and present an error instead.

Fixes #607

### DIFF
--- a/pkg/prototype/flags.go
+++ b/pkg/prototype/flags.go
@@ -1,0 +1,55 @@
+// Copyright 2018 The ksonnet authors
+//
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package prototype
+
+import (
+	"fmt"
+
+	"github.com/spf13/pflag"
+)
+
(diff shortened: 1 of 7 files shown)


pkg/prototype/flags.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// BindFlags creates a flag set using a prototype's parameters.
+func BindFlags(p *Prototype) (fs *pflag.FlagSet, err error) {
+	fs = pflag.NewFlagSet("preview", pflag.ContinueOnError)
+
+	fs.String("module", "", "Component module")
+
+	for _, param := range p.RequiredParams() {
+		defer func() {
```
shomron: This pushes a bunch of defers onto the stack?

pkg/prototype/flags.go
```diff
@@ -0,0 +1,59 @@
...
+			if r := recover(); r != nil {
+				err = &FlagDefinitionError{name: p.Name}
```
shomron: The param name is `param.Name`, not `p.Name`.

pkg/prototype/flags.go
```diff
@@ -0,0 +1,59 @@
...
+			}
+		}()
+		fs.String(param.Name, "", param.Description)
```
shomron: I don't like depending on panic/recover for this - we can detect collisions by asking the FlagSet:

```go
	for _, param := range p.RequiredParams() {
		// Avoid flag collisions
		if fs.Lookup(param.Name) != nil {
			return nil, &FlagDefinitionError{name: param.Name}
		}

		fs.String(param.Name, "", param.Description)
	}
```

PR #611, opened 2018-06-13
fixes to CacheRoot, additional tests for GitHub registry.

shomron:
also: rename registry.makePath -> regsitry.registrySpecFilePath

### DIFF
--- a/pkg/registry/github_test.go
+++ b/pkg/registry/github_test.go
@@ -34,7 +34,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func makeGh(t *testing.T, u, sha1 string) (*GitHub, *mocks.GitHub) {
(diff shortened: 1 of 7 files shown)


pkg/registry/github_test.go
```diff
@@ -508,3 +511,71 @@ func Test_parseGitHubURI(t *testing.T) {
...
+			path:     "registry.yaml",
+			expected: "incubator/registry.yaml",
+		},
+		{
+			name:     "registry name and url tail are different",
+			url:      "github.com/ksonnet/parts/tree/master/foobar",
+			path:     "foobar/apache/parts.yaml",
+			expected: "incubator/apache/parts.yaml",
+		},
+		// {
```
bryanl: Will these work at a later date?

shomron: Only if I bypass the constructor. Is that worth the bother?

PR #612, opened 2018-06-14
enable env set to change server

underrun:
adds --server flag to `ks env set` options

also also fixes a couple bugs and adds version to INFO log on init

### DIFF
--- a/pkg/clicmd/env_set.go
+++ b/pkg/clicmd/env_set.go
@@ -26,6 +26,7 @@ import (
 const (
 	vEnvSetName      = "env-set-name"
 	vEnvSetNamespace = "env-set-namespace"
+	vEnvSetServer    = "env-set-server"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/clicmd/env_set.go
```diff
@@ -41,6 +42,7 @@ var envSetCmd = &cobra.Command{
 			actions.OptionEnvName:    args[0],
 			actions.OptionNewEnvName: viper.GetString(vEnvSetName),
 			actions.OptionNamespace:  viper.GetString(vEnvSetNamespace),
+			actions.OptionServer:     viper.GetString(vEnvSetServer),
```
bryanl: If you add a new option, you'll have to adjust the test to expect the new option.

underrun: add tests for both clicmd and actions

PR #620, opened 2018-06-18
Use factory functions for creating cobra commands

bryanl:
Remove global commands, afero filesystem, and ksonnet app object. This makes is easier to test the commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ _testmain.go
 
 # gotags
 tags
+
+.DS_Store
--- a/pkg/clicmd/registry_add.go
+++ b/pkg/clicmd/registry_add.go
@@ -21,6 +21,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/ksonnet/ksonnet/pkg/actions"
+	"github.com/ksonnet/ksonnet/pkg/app"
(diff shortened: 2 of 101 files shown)


pkg/clicmd/registry_add.go
```diff
@@ -77,22 +60,44 @@ SCM configuration.
...
-ks registry add databases github.com/example/tree/master/reg --version=0.0.1`,
-}
+ks registry add databases github.com/example/tree/master/reg --version=0.0.1`
+)
 
-func init() {
-	registryCmd.AddCommand(registryAddCmd)
+func newRegistryAddCmd(a app.App) *cobra.Command {
+	registryAddCmd := &cobra.Command{
+		Use:     "add <registry-name> <registry-uri>",
```
underrun: probably not worth the work it would take, but it could potentially be nice to have the usage text pulled out along side the descriptions and examples so that if something wants the usage it doesn't have to create a new cobra.Command to get it.

.gitignore
```diff
@@ -31,3 +31,5 @@ _testmain.go
 
 # gotags
 tags
+
+.DS_Store
```
underrun: picky editors like newlines :-)

PR #636, opened 2018-06-20
Add generators for all mocks

### LINKED ISSUE #605: Generate mocks with go generation tags
Add go generate tags to all generated mocks

e.g. in `pkg/app/mocks/generate.go`

```go
//go:generate mockery -dir .. -output . -name=App
```

bryanl:
Add generators for mocks in the following packages:

pkg/cluster/mocks
pkg/app/mocks
pkg/util/github/mocks
pkg/component/mocks
pkg/registry/mocks
pkg/pkg/mocks

Fixes #605


pkg/prototype/rice-box.go
```diff
@@ -25,34 +10,34 @@ func init() {
 	// define files
 	file2 := &embedded.EmbeddedFile{
 		Filename:    "config-map.jsonnet",
-		FileModTime: time.Unix(1526913293, 0),
+		FileModTime: time.Unix(1526918148, 0),
```
shomron: Did these changes sneak in?

pkg/prototype/rice-box.go
```diff
@@ -1,18 +1,3 @@
-// Copyright 2018 The ksonnet authors
```
shomron: Keeping the copyright notice on generated files has been somewhat a pain point.

shomron: It looks like it's at least addressable with _mockery_ - something to look at later:

```
  -note string
    	comment to insert into prologue of each generated file
```

pkg/prototype/rice-box.go
bryanl: I ran generate and it updated the mtime. I didn't see it as a problem

Issue #427, opened 2018-04-08
ksonnet's kubernetes version resolver needs to be expanded to work with multiple distributions of Kubernetes


fxue:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
 /kind bug
>
> /kind feature

**What happened**:
Found this while resolving Kubeflow [issue](https://github.com/kubeflow/kubeflow/issues/521).
copying it here.

**What you expected to happen**:
correct k8s picked up

**How to reproduce it (as minimally and precisely as possible)**:

```
➜  ks version
ksonnet version: 0.9.2
jsonnet version: v0.9.5
client-go version: 1.8
```
getting a different 404 error now. 
```
➜  ks show ${KF_ENV} -c kubeflow-core
ERROR Recieved status code '404' when trying to retrieve OpenAPI schema for cluster version 'v3.9.14' from URL 'https://raw.githubusercontent.com/kubernetes/kubernetes/v3.9.14/api/openapi-spec/swagger.json'

```
seems like the problem is with using the openshift version instead of k8s version
```
➜ oc version
oc v3.9.7
kubernetes v1.9.1+a0ce1bc657
features: Basic-Auth

Server https://api.starter-us-west-1.openshift.com:443
openshift v3.9.14
kubernetes v1.9.1+a0ce1bc657
```

**Anything else we need to know?**:

**Environment**:
- ksonnet version (use `ks version`):0.9.2
- Kubernetes version (use `kubectl version`):v1.9.1

pdmack:
This is due to the fact that the ks client retrieves the version from the Kubernetes master swagger.json endpoint. However, for OpenShift this value is its own (e.g., v3.9.14). Once the version is parsed it's assumed to be registered here:
```
metadata/lib/clusterspec.go:    k8sVersionURLTemplate = "https://raw.githubusercontent.com/kubernetes/kubernetes/%s/api/openapi-sp
metadata/lib/clusterspec.go:    versionURL := fmt.Sprintf(k8sVersionURLTemplate, cs.k8sVersion)
``` 
which wouldn't apply for OpenShift.

bryanl:
In ksonnet 0.10, the logic for finding the schema has changed. If ksonnet can read swagger.json, it'll use that file. There is a possibility that will backported to 0.9, but that won't happen until after 0.10 is released.

mattf:
[...]

PR #640, opened 2018-06-23
Shift to ServerVersion call to get version

underrun:
instead of using OpenAPISchema for version because openshift stores its
own version data in Info.Version

this only addresses the app init section - there are other places where version info could be gathered for environments that will need to be addressed as well. before i head down that road i'm pushing this to check and see if this is the right direction to move in - which is why it's WIP right now.

related to #427

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strings"
 
 	"github.com/ksonnet/ksonnet/pkg/app"
 	str "github.com/ksonnet/ksonnet/pkg/util/strings"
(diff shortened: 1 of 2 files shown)


bryanl:
This won't be reviewed until the `wip` is removed.

underrun:
@bryanl the reason it didn't work against GKE is because the git version representation was different than on openshift so rather than splitting on anything we can look for something that looks like a kubernetes version instead.

This also does hit the cluster to ask for version if at all possible. if it fails to hit the live cluster we fall back to 1.8.0 still. The major difference with this PR is that we are hitting the api at `/version` rather than `/openapi/v2` - openshift doesn't include the kubernetes version in the swagger.json - it's only in the `/version` end point.

Also because `/version` has major and minor version but not patch, I can't just render the full version from parts - I need to parse the git version which includes commit info beyond the latest k8s tag (but does have `major.minor.patch` version info)

PR #643, opened 2018-06-24
param set should understand module dot paths

### LINKED ISSUE #642: Setting parameters in modules doesn’t understand dot notation
/kind bug

Modules are identified with a dot path: `apps.name.component`. When setting parameters, you be able to use the dot path: `ks param set apps.name.component field value`.

This doesn't work, but using slashes does.

bryanl:
Fixes #642

### DIFF
--- a/pkg/component/manager.go
+++ b/pkg/component/manager.go
@@ -24,37 +24,38 @@ import (
 	"github.com/spf13/afero"
 )
 
+// ResolvePath resolves a given path to a module and a component.
 func ResolvePath(ksApp app.App, path string) (Module, Component, error) {
 	isDir, err := isComponentDir2(ksApp, path)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "check for namespace directory")
+		return nil, nil, errors.Wrap(err, "checking for module directory")
 	}
 
 	if isDir {
-		ns, err := GetModule(ksApp, path)
+		m, err := GetModule(ksApp, path)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.Wrapf(err, "retrieving module %s", path)
 		}
 
-		return ns, nil, nil
+		return m, nil, nil
 	}
 
-	module, cName, err := checkComponent(ksApp, path)
+	moduleName, componentName, err := extractPathParts(ksApp, path)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, errors.Wrapf(err, "extracting module and component names from %s", path)
 	}
 
-	ns, err := GetModule(ksApp, module)
+	m, err := GetModule(ksApp, moduleName)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, errors.Wrapf(err, "retrieving module %s", path)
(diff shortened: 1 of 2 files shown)


pkg/component/manager.go
```diff
@@ -24,37 +24,38 @@ import (
...
 
-		return ns, nil, nil
+		return m, nil, nil
 	}
 
-	module, cName, err := checkComponent(ksApp, path)
+	moduleName, componentName, err := extractPathParts(ksApp, path)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, errors.Wrapf(err, "checking component %s", path)
```
underrun: "extracting path parts" instead of checking component to match new function name

pkg/component/manager.go
```diff
@@ -24,37 +24,38 @@ import (
...
-			return nil, nil, err
+			return nil, nil, errors.Wrapf(err, "retrieving module %s", path)
 		}
 
-		return ns, nil, nil
+		return m, nil, nil
 	}
 
-	module, cName, err := checkComponent(ksApp, path)
+	moduleName, componentName, err := extractPathParts(ksApp, path)
```
underrun: we're passing `path` in as `name` to extractPathParts ... if we've gotten here then `path` is not a path. previous to here `path` could actually be a path (if isDir is true it's an actual path right?) ... i feel like `name` feels better all the way up on this where `name` could be a path and you early out if it is.

pkg/component/manager_test.go
```diff
@@ -52,29 +52,32 @@ func Test_ResolvePath(t *testing.T) {
...
-			isErr  bool
+			name              string
+			expectedComponent string
+			expectedModule    string
+			isErr             bool
 		}{
 			{
-				name:   "/",
-				module: "/",
+				name:           "/",
```
underrun: so .. this works, but contextually being able to specify a path with / in it to ResolvePath but also that names cannot have / in them feels a bit rough to me - can we go all dots?

pkg/component/manager_test.go
```diff
@@ -52,29 +52,32 @@ func Test_ResolvePath(t *testing.T) {
...
-				cName:  "deployment",
+				name:              "nested.deployment",
+				expectedModule:    "nested",
+				expectedComponent: "deployment",
 			},
 			{
-				name:   "nested/deployment",
-				module: "nested",
-				cName:  "deployment",
+				name:           "nested",
```
underrun: this test is not nested and the case is also covered in "deployment" we should remove this test

pkg/component/manager.go
bryanl: I replaced it with "extracting module and component names from %s"

pkg/component/manager_test.go
bryanl: I thought about all dots, but having one named "." felt weird to me, so I left it as "/".

pkg/component/manager.go
bryanl: It's a path and not a name. I updated the function.

PR #649, opened 2018-06-25
Use values file for prototype use/preview

### LINKED ISSUE #550: Create values flag to supply values to `ks prototype use`

bryanl:
This change introduces a values file to be used when setting prototype
values for the use and preview commands.

The format of the file is Jsonnet that returns an object.

Notes:

* if values are added to the file and aren't consumed by prototype, they are ignored.

Fixes #550 

### DIFF
--- a/docs/cli-reference/ks_prototype_preview.md
+++ b/docs/cli-reference/ks_prototype_preview.md
@@ -33,6 +33,20 @@ ks prototype preview single-port-deployment \
   --name=nginx                              \
   --image=nginx                             \
   --port=80
+
+# Preview prototype 'io.ksonnet.pkg.single-port-deployment', using the
+# 'nginx' image, and port 80 exposed with a values file.
+ks prototype preview simple-port-deployment \
+  --name=nginx                              \
+  --values-file=ks-values
+
+Where 'ks-values' is a jsonnet file with the contents:
+
+{
+	image: "nginx",
+	port: 80,
+}
+
 ```
 
 ### Options
--- a/pkg/prototype/flags.go
+++ b/pkg/prototype/flags.go
@@ -18,6 +18,9 @@ package prototype
 import (
 	"fmt"
 
+	"github.com/ksonnet/ksonnet/pkg/util/strings"
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
 	"github.com/spf13/pflag"
 )
 
@@ -35,6 +38,7 @@ func (e *FlagDefinitionError) Error() string {
 func BindFlags(p *Prototype) (fs *pflag.FlagSet, err error) {
 	fs = pflag.NewFlagSet("prototype-flags", pflag.ContinueOnError)
 
+	fs.String("values-file", "", "Prototype values file (file returns a Jsonnet object)")
 	fs.String("module", "", "Component module")
 
 	for _, param := range p.RequiredParams() {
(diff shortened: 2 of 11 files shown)


pkg/prototype/flags_test.go
```diff
@@ -109,3 +112,194 @@ func TestBindFlags_duplicate_optional_param(t *testing.T) {
...
+		{
+			name: "flag not defined",
+			p:    validPrototype,
+			initFlags: func(t *testing.T, p *Prototype, args []string) *pflag.FlagSet {
+				flags := pflag.NewFlagSet("prototype-flags", pflag.ContinueOnError)
+				return flags
+			},
+			args: []string{
+				"--name=name",
+				"--val=a",
```
underrun: a name like `--undefined-flag` indicates where the error in this test will be better than `--val` here

pkg/prototype/flags_test.go
```diff
@@ -109,3 +112,194 @@ func TestBindFlags_duplicate_optional_param(t *testing.T) {
...
+				flags, err := BindFlags(p)
+				require.NoError(t, err)
+
+				err = flags.Parse(args)
+				require.NoError(t, err)
+
+				return flags
+			},
+			args: []string{
+				"--values-file=/values-file",
```
underrun: maybe `/missing-values-file`

pkg/prototype/values_file.go
```diff
@@ -0,0 +1,105 @@
...
+
+	v, err := vm.EvaluateSnippet("getValue", snippetFieldValue)
+	if err != nil {
+		return "", errors.Errorf("key %q was not found", k)
+	}
+
+	return strings.TrimSpace(v), nil
+}
+
+var snippetFieldValue = `
```
underrun: does this need to be outside the function `Get`?

docs/cli-reference/ks_prototype_preview.md
```diff
@@ -33,6 +33,20 @@ ks prototype preview single-port-deployment \
...
   --image=nginx                             \
   --port=80
+
+# Preview prototype 'io.ksonnet.pkg.single-port-deployment', using the
+# 'nginx' image, and port 80 exposed with a values file.
+ks prototype preview simple-port-deployment \
+  --name=nginx                              \
+  --values-file=ks-values
+
+Where 'ks-values' is a jsonnet file with the contents:
```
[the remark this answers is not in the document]
bryanl: Jsonnet should cover all the cases. Why would you want to use YAML?

pkg/prototype/flags.go
```diff
@@ -53,3 +57,102 @@ func BindFlags(p *Prototype) (fs *pflag.FlagSet, err error) {
...
+		}
+	}
+
+	valuesFilePath, err := flags.GetString("values-file")
+	if err != nil {
+		return nil, errors.Wrap(err, "finding values file flag")
+	}
+
+	if valuesFilePath != "" {
+		readFromValuesFile(fs, values, valuesFilePath)
```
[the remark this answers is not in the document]
bryanl: I wasn't going for a functional approach in this case.

pkg/prototype/values_file.go
bryanl: What are the reasons for moving it?

bryanl:
Thanks @underrun

PR #653, opened 2018-06-25
Add modules to concepts.md

GuessWhoSamFoo:
Adding a figure can also be helpful although modules are not that abstract.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -8,6 +8,7 @@ The ksonnet framework is centered around a couple of key concepts and terms. Thi
   * [Component](#component)
   * [Prototype](#prototype)
   * [Parameter](#parameter)
+  * [Module](#module)
   * [Part](#part)
   * [Package](#package)
   * [Registry](#registry)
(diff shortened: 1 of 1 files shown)


docs/concepts.md
```diff
@@ -126,6 +127,22 @@ For example, you can use params to ensure that you have 3 Redis replicas in your
...
+
+Modules provide a way for you to share components across environments. More concisely, a module refers to a subdirectory in `components/` containing its own `params.libsonnet`.
+
+For example, `ks module create <module-name>` creates `components/<module-name>/params.libsonnet`.
+
+ksonnet will use the components defined in `components/` by default. If you want to use another set of components, modules provides a way to do this without initializing another ksonnet app.
+
+A module can:
+
+* used across multiple [environments](#environment) similar to components.
```
bryanl: Would this look better if `be` was prepended?

docs/concepts.md
```diff
@@ -126,6 +127,22 @@ For example, you can use params to ensure that you have 3 Redis replicas in your
 
 ---
 
+### Module
```
bryanl: Could you add something about making your configuration more modular?

docs/concepts.md
GuessWhoSamFoo: Fixed

PR #661, opened 2018-06-27
Support environment arguments in module parameters

### LINKED ISSUE #579: __ksonnet/environments should be available in params.libsonnet
`__ksonnet/envrionments` ext var should be available during parameter processing

/cc @jessesuen

bryanl:
It is possible that a user might want to use environment parameters
(namespace or server) in their parameters. This would not work with
ksonnet for two reasons:

1. ksonnet didn't provide external variables with those settings
1. params.libsonnet was evaluated as an object

This change adds support for these two items.

```
bash-3.2$ ks param list
COMPONENT PARAM         VALUE
========= =====         =====
ds        arr           [1, 2, 3, 4]
ds        containerPort 80
ds        image         'gcr.io/heptio-images/ks-guestbook-demo:0.1'
ds        name          'ds'
ds        namespace     ''
ds        obj           { a: 'b' }
ds        replicas      1
ds        server        ''
ds        servicePort   80
ds        type          'ClusterIP'
bash-3.2$ ks param list --env default
COMPONENT PARAM         VALUE
========= =====         =====
ds        arr           [1, 2, 3, 4]
ds        containerPort 80
ds        image         'gcr.io/heptio-images/ks-guestbook-demo:0.1'
ds        name          'ds'
ds        namespace     'default'
ds        obj           { a: 'b' }
ds        replicas      1
ds        server        'https://localhost:6443'
ds        servicePort   80
ds        type          'ClusterIP'
```

An example parameters file that produces the preceeding output:

```js
local env = std.extVar("__ksonnet/environments");

{
  global: {
  },
  components: {
    ds: {
      containerPort: 80,
      image: "gcr.io/heptio-images/ks-guestbook-demo:0.1",
      name: "ds",
      replicas: 1,
      servicePort: 80,
      type: "ClusterIP",
      namespace: env.namespace,
      server: env.server,
      obj: {a:"b"},
      arr: [1,2,3,4],
    },
  },
}
```

* note: params files aren't currently generated with the `env` local.

Fixes #579 


pkg/actions/param_list.go
```diff
@@ -62,10 +72,17 @@ func NewParamList(m map[string]interface{}) (*ParamList, error) {
...
-		findModulesFn:   component.ModulesFromEnv,
-		findModuleFn:    component.GetModule,
-		findComponentFn: component.LocateComponent,
+		out:          os.Stdout,
+		findModuleFn: component.GetModule,
+	}
+
+	if pl.app != nil {
+		p := pipeline.New(pl.app, pl.envName)
+		pl.modulesFn = p.Modules
```
shomron: We treat app as optional here by not returning an error, but depend on `pl.modulesFn`, `pl.envParametersFn`, and `pl.lister` being non-nil below.

pkg/component/module.go
```diff
@@ -217,15 +240,38 @@ func (mp *ModuleParameter) IsSameType(other ModuleParameter) bool {
...
-	object, err := jsonnet.Parse("params.libsonnet", s)
+	envCode, err := params.JsonnetEnvObject(m.app, envName)
 	if err != nil {
-		return "", err
+		return "", errors.Wrap(err, "building environment argument")
+	}
+
+	vm := jsonnet.NewVM()
+	vm.AddJPath(
+		filepath.Join(m.app.Root(), "vendor"),
```
shomron: Are there any circumstances where we would need the _revendored_ (versioned) packages available here?

pkg/params/lister.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		entries = append(entries, paramEntries...)
+	}
+
+	sort.Slice(entries, func(i, j int) bool {
+		if entries[i].ComponentName < entries[j].ComponentName {
+			return true
+		}
+
```
shomron: Missing complementary comparison

```
if entries[i].ComponentName > entries[j].ComponentName {
    return false
}
```

pkg/params/lister.go
```diff
@@ -0,0 +1,280 @@
...
+	return &Lister{
+		AppRoot:     appRoot,
+		Destination: destination,
+
+		createEntry: ec.Create,
+	}
+}
+
+// List lists parameters in a slice of Entry.
+func (l *Lister) List(r io.Reader, componentName string) ([]Entry, error) {
```
shomron: A comment on the expected input format (example jsonnet) would be useful for the uninitiated.

pkg/params/lister.go
```diff
@@ -0,0 +1,280 @@
...
+
+		return entries[i].ParamName < entries[j].ParamName
+	})
+
+	return entries, nil
+}
+
+// componentsObject finds the components object in side of a Jsonnet object in params format.
+func (l *Lister) componentsObject(object *astext.Object) (*astext.Object, error) {
+	for i := range object.Fields {
```
shomron: This is not significant, but I was wondering why not use the two valued variant of `range`?

pkg/params/lister.go
```diff
@@ -0,0 +1,280 @@
...
+func (l *Lister) buildObject(source string) (*astext.Object, error) {
+	// TODO: this code is repeated in module.Resolveparams, and should be centralized.
+	envCode, err := l.destinationObject()
+	if err != nil {
+		return nil, errors.Wrap(err, "building environment object")
+	}
+
+	vm := jsonnet.NewVM()
+	vm.AddJPath(
+		filepath.Join(l.AppRoot, "vendor"),
```
shomron: Same question about revendored packages.

pkg/actions/param_list.go
bryanl: Yes, but the `ol.err` check below will bail out if app is not set. So the assumption is if you got this far, everything was set properly.

pkg/component/module.go
bryanl: Is there currently an API for determining this?

pkg/params/lister.go
bryanl: I'm not sure I understand. This is a less function. Why would you need a complementary comparison?

pkg/component/module.go
shomron: There's a non-exported function called `revendorPackages` which does revendoring and returns a path to import. But before going down that path, I was asking if the parameter resolution we're trying to do here would rely on installed packages?

pkg/actions/param_list.go
shomron: In that case maybe reordering to check `ol.err` before accessing `pl.app` here would be clearer?

pkg/component/module.go
bryanl: It knows nothing about packages. It only looks in vendor.

pkg/params/lister.go
bryanl: Because I want to copy the field out rather than reusing the variable.

pkg/params/lister.go
shomron: Because it's comparing on two keys.

Currently the following would output in the same order instead of the expected:

```
{ComponentName: "b", ParamName: "1"},
{ComponentName: "a", ParamName: "2"},
```

shomron: ...that is, you're not differentiating between keyA equality (need to check keyB) and keyA being out-of-order (should not consult keyB).

pkg/component/module.go
shomron: Let me rephrase the question - would the jsonnet in `params.jsonnet` depend on code in vendored packages? If not, why do we include `vendor` in the JPath? If yes, it sounds like we should support the revendoring logic so that `params.jsonnet` can refer to unversioned package imports.

pkg/params/lister.go
shomron: The variable is reused, but isn't the map value being copied over it each iteration?
I had trouble finding the justification in the spec, but that was my understanding.

https://play.golang.org/p/Cp5Wb6MliLe

It's possible I'm misunderstanding.

pkg/component/module.go
shomron: Regardless of the answer, we can address this separately if needed.

PR #668, opened 2018-06-28
Create prototypes for helm packages

### LINKED ISSUE #664: Helm based packages generates a prototype
/feature

Generate a prototype for a helm chart

```
ks generate helm-name name <options>
```

This generates a component that makes the following call:

```
std.prune(std.native("renderHelmChart")("repository", "version", "version", params))
```

Where:

* repository: repository name
* chart: chart name
* version: chart version or empty string for latest
* params: object with params

bryanl:
Adds:

* prototype for packages from helm registries
* support for helm registries in `ks registry add`

Detail:

* add a registry
`ks registry add helm-stable https://kubernetes-charts.storage.googleapis.com`

* install helm charts as packages
`ks pkg install helm-stable/redis`

* Generates a component:
`ks generate helm-stable-redis hc1`

Helm values can be overridden with the generated component's `values` key. 

```js
local env = std.extVar("__ksonnet/environments");
local params = std.extVar("__ksonnet/params").components["hc1"];

std.prune(std.native("renderHelmChart")(
        // registry name
        "helm-stable",
        // chart name
        "redis",
        // chart version
        params.version,
        // chart values overrides
        params.values,
        // component name
        params.name,
 ))
```

Notes:

* Component name will be used as the helm release name by default. This can be configured with params.

Fixes #664


pkg/helm/renderer.go
```diff
@@ -0,0 +1,237 @@
...
+	return rendered, nil
+}
+
+func checkDependencies(ch *chart.Chart, reqs *chartutil.Requirements) error {
+	missing := []string{}
+
+	deps := ch.GetDependencies()
+	for _, r := range reqs.Dependencies {
+		found := false
+		for _, d := range deps {
```
shomron: What are the numbers we're dealing with here? Can we reimplement to be less O(n*m) as a followup issue?

pkg/helm/renderer.go
```diff
@@ -0,0 +1,237 @@
...
+			if d.Metadata.Name == r.Name {
```
shomron: `Chart.Metadata` is a pointer - should we guard against nil Metadata?

pkg/helm/renderer.go
```diff
@@ -0,0 +1,237 @@
...
+
+	if len(missing) > 0 {
+		return fmt.Errorf("found in requirements.yaml, but missing in charts/ directory: %s", strings.Join(missing, ", "))
+	}
+	return nil
+}
+
+// LatestChartVersion finds the latest vendored version of a chart.
+func LatestChartVersion(a app.App, repoName, chartName string) (string, error) {
+	path := filepath.Join(a.Root(), "vendor", repoName, chartName, "helm")
```
shomron: I was going to suggest using Using pkg.Helm.Path(), but that would be recursive. Maybe we can refactor that a bit.

pkg/helm/renderer.go
```diff
@@ -0,0 +1,237 @@
...
+
+	fis, err := afero.ReadDir(a.Fs(), path)
```
shomron: Looking at this again, I feel the package manger should be supplying the installed packages with their versions - here we could just sort them. This can be a followup issue.

pkg/helm/renderer.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	if len(missing) > 0 {
+		return fmt.Errorf("found in requirements.yaml, but missing in charts/ directory: %s", strings.Join(missing, ", "))
+	}
+	return nil
+}
+
+// LatestChartVersion finds the latest vendored version of a chart.
+func LatestChartVersion(a app.App, repoName, chartName string) (string, error) {
```
shomron: `renderer.go` is not the first place I'd go looking for this function. Why was it moved from `helm.go`?

pkg/util/test/test.go
```diff
@@ -63,16 +63,16 @@ func StageFile(t *testing.T, fs afero.Fs, src, dest string) {
 // testdata.
 func StageDir(t *testing.T, fs afero.Fs, src, dest string) {
 	root, err := filepath.Abs(filepath.Join("testdata", src))
-	require.NoError(t, err)
+	require.NoError(t, err, "finding absolution path")
```
shomron: Finding absolution is important, but I think you meant absolute 😃

pkg/helm/renderer.go
bryanl: If it ever approached five, I'd be surprised.

pkg/helm/renderer.go
```diff
@@ -0,0 +1,237 @@
...
+		if !ok {
+			return nil, errors.New("invalid Helm chart values")
+		}
+
+		componentName, ok := input[4].(string)
+		if !ok {
+			return nil, errors.New("invalid component name")
+		}
+
+		return r.Render(repoName, chartName, chartVersion, componentName, values)
```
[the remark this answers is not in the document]
bryanl: I'm not confident at all. The generated bit calls the jsonnet function. This could be changed anytime.

Issue #660, opened 2018-06-27
ks pkg install - side-by-side support for versioned packages


shomron:
`ks pkg install` should allow installing multiple versions of a package side-by-side in an application.

Current behavior detects collisions based on name only and rejects a newer (or older) version of a package if any other version is already installed.

Part of #629

PR #669, opened 2018-06-28
pkg install honors versions when checking for conflicts

shomron:
This commit allows installing a package when a different version is
already installed. Currently, if the package is installed globally (the
only option) - it will overwrite the previous version reference.
A followup change will allow installing into environments to work around
this limitation.

Part of #660

Also:

* Add PackageManager.IsInstalled - fuzzy-match installation check for packages, without parsing their manifests


shomron:
Exercising the change:

```
ks init multi && cd multi

ks --verbose=5 pkg install incubator/mysql
grep -A4 libraries app.yaml
tree vendor

ks --verbose=5 registry set incubator --uri https://github.com/shomron/parts/tree/nested-registry/nested/registry/incubator

# Overwrites global reference, installs new version
ks --verbose=5 pkg install incubator/mysql
grep -A4 libraries app.yaml
tree vendor

# Same version, show conflict
ks --verbose=5 pkg install incubator/mysql
```

pkg/registry/cache_test.go
```diff
@@ -59,9 +67,16 @@ func Test_CacheDependency(t *testing.T) {
 		}
 		for _, lib := range libs {
 			a.On("UpdateLib", lib.Name, lib).Return(nil)
+
+			var checker installedChecker
+			// checker.On("IsInstalled", pkg.Descriptor{
```
bryanl: Is this coming back in the future, or should it be deleted?

bryanl:
Will approve after the unused code is deleted.

pkg/registry/cache_test.go
shomron: The test may become more elaborate in the future - for now I'll take these out. 👍

shomron:
Removed unused code.

PR #671, opened 2018-06-28
ks version and -h flag should not findRoot

### LINKED ISSUE #659: Enable ks sub-command -h to run outside a ksonnet app
Running `ks param -h` outside a ksonnet app directory hard fails with:
```
ERRO[0000] unable create ksonnet command                 error="unable to find ksonnet project"
```
This is a regression since 0.11.0 (broke with 8b152611)

GuessWhoSamFoo:
Allow commands with `-h` flag and `ks version` to run outside of a ks application.

Fixes #659 


bryanl:
Hey @GuessWhoSamFoo:

The cases that shouldn't look for a root are running ks with `init`, 'version`, or if the --help flag is passed to ks directly. Perhaps there could be a function that checks for that and skips accordingly because this current change checks for `init` at the top, and then in conditional checks for the others.

shomron:
@GuessWhoSamFoo  please look at this in context of work done in #679.

bryanl:
I think you all should combine forces and pick this or #679. Either way there is going to be a merge conflict, so might as well avoid it.

GuessWhoSamFoo:
Three cases:

ks init - `init.go` calls app.Load and skips finding root
ks version - addressed above to use preparser
help flag - allow app.App to be nil and implicitly not require finding app root

pkg/clicmd/root.go
```diff
@@ -106,8 +106,8 @@ func NewRoot(appFs afero.Fs, wd string, args []string) (*cobra.Command, error) {
 		return nil, err
 	}
 
-	if len(args) > 0 && cmdName != "init" {
-		a, err = app.Load(appFs, wd, false)
+	if cmdName == "version" {
```
shomron: Where will non-`version` commands have their app loaded if not here?

GuessWhoSamFoo: Disregard; something weird happened with having VS code and vim opening the same file

pkg/actions/actions.go
```diff
@@ -325,7 +324,7 @@ func (o *optionLoader) LoadClientConfig() *client.Config {
 func (o *optionLoader) LoadApp() app.App {
 	i := o.load(OptionApp)
 	if i == nil {
-		o.err = ErrNotInApp
+		o.err = ErrInvalidApp
```
shomron: This means we lose visibility into the root cause -  is this preferable to `errors.Wrap(o.err, ...)`?

pkg/clicmd/root.go
```diff
@@ -106,7 +117,8 @@ func NewRoot(appFs afero.Fs, wd string, args []string) (*cobra.Command, error) {
 		return nil, err
 	}
 
-	if len(args) > 0 && cmdName != "init" {
+	cmds := []string{"init", "version"}
+	if len(args) > 0 && !(strings.InSlice(cmdName, cmds) || containsHelp(args)) {
```
shomron: This is getting hard to read.
It's a matter of taste, but my preference would be something like this:

```go
switch {
case string.InSlice(cmdName, cmds):
    // commands that should not load an app
case containsHelp(args):
    // help does not require an app
case cmdName != "":
    a, err = app.Load(appFs, wd, false)
    ...
default:
    // noop
}
```

pkg/clicmd/root.go
shomron: This has bit me in the past - if you're using a neovim variant, you may need to add the following to your ~/.vimrc to get it to auto-reload external changes when it receives focus:

```
" autoload external changes
set autoread
au FocusGained * :checktime
```

pkg/actions/actions.go
```diff
@@ -325,7 +324,7 @@ func (o *optionLoader) LoadClientConfig() *client.Config {
 func (o *optionLoader) LoadApp() app.App {
 	i := o.load(OptionApp)
 	if i == nil {
-		o.err = ErrNotInApp
+		o.err = errors.Wrap(ErrInvalidApp, "found app.yaml but cannot load app")
```
shomron: The other way around - you'd need to wrap `o.err` with your context, so assuming have extra context that is useful it might look something like:

```
o.err = errors.Wrap(o.err, "extra context about what you were doing")
```

In this case it might be just as well to leave the original error be.

pkg/clicmd/root.go
```diff
@@ -80,18 +81,25 @@ func appRoot() (string, error) {
...
-		return "", err
+		return "", false, err
 	}
 	if len(fset.Args()) == 0 {
-		return "", nil
+		return "", false, nil
 	}
 
-	return fset.Args()[0], nil
+	flags := []string{"-h", "--help"}
```
shomron: This is not necessary - pflag is already parsing the command line.
The `BoolP` method call above returns a `*bool` that points to a boolean representing the help flag. You can capture that and just return its underlying value to the caller.

pkg/clicmd/root.go
```diff
@@ -80,18 +81,25 @@ func appRoot() (string, error) {
 
 // parseCommand does an early parse of the command line and returns
 // what will ultimately be recognized as the command by cobra.
```
shomron: Please update the comment to document the new return value.

pkg/clicmd/root.go
```diff
@@ -122,17 +131,24 @@ func NewRoot(appFs afero.Fs, wd string, args []string) (*cobra.Command, error) {
...
+	cmdName, hasHelpFlag, err := parseCommand(args)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(args) > 0 && cmdName != "init" {
+	cmds := []string{"init", "version", "help"}
+	switch {
+	// Commands that do not require a ksonnet application
+	case strings.InSlice(cmdName, cmds) || hasHelpFlag:
```
shomron: `case` supports a list of cases which are effectively ORed, so another way to say the same thing might be:

```
case strings.InSlice(cmdName, cmds), hasHelpFlag:
```

Either way is fine.
See: https://golang.org/doc/effective_go.html?#switch

pkg/clicmd/root_test.go
```diff
@@ -46,7 +46,7 @@ func Test_parseCommand(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		cmd, err := parseCommand(tc.args)
+		cmd, _, err := parseCommand(tc.args)
```
shomron: Better to add tests for the new functionality than to ignore it.

pkg/actions/actions.go
GuessWhoSamFoo: Switched back to original error

pkg/clicmd/root.go
GuessWhoSamFoo: Removed superfluous code and used BoolP

pkg/clicmd/root.go
GuessWhoSamFoo: Updated comments on return value of parseCommand to include a boolean

pkg/clicmd/root_test.go
GuessWhoSamFoo: Updated tests for new parseCommand return vals

pkg/clicmd/root_test.go
```diff
@@ -27,28 +27,44 @@ func Test_parseCommand(t *testing.T) {
...
+			helpFlag: false,
 		},
 	}
 
 	for _, tc := range tests {
-		cmd, err := parseCommand(tc.args)
+		cmd, hasHelpFlag, err := parseCommand(tc.args)
 		require.NoError(t, err)
-		assert.Equal(t, tc.expected, cmd, tc.name)
+		assert.Equal(t, tc.expected, cmd, tc.name, hasHelpFlag, tc.helpFlag)
```
bryanl: The fourth argument for `assert.Equal` is a printf format string. The fifth and beyond are arguments for the format flag.

bryanl: What about the case `ks -h` or `ks --help`, currently that prints `ERRO[0000] unable create ksonnet command`

Issue #670, opened 2018-06-28
Why does string representation matter depending on whether its imported from libsonnet file?


jlewi:
**What happened**:

I define the same service two different ways in the following two files

1. I define it directly in the jsonnet file
1. I define it an import it from the libsonnet file

https://github.com/jlewi/kubeflow-dev/blob/0a78fee4daace5a7adfcbf5efecb860a0bfd1f00/kubeflow-deployments/ks-app/components/test-central.jsonnet#L1
https://github.com/jlewi/kubeflow-dev/blob/0a78fee4daace5a7adfcbf5efecb860a0bfd1f00/kubeflow-deployments/ks-app/components/test-central.libsonnet#L1

The resulting service is different. In the case where it is defined in libsonnet

```
  annotations:
    getambassador.io/config: |-
      ---
      apiVersion: ambassador/v0
      kind:  Mapping
      name: centralui-mapping
      prefix: /
      rewrite: /
      service: centraldashboard.kubeflow
```

When its defined in the jsonnet file

```
  annotations:
    getambassador.io/config: '---\napiVersion: ambassador/v0\nkind:  Mapping\nname:
      uiLocal-mapping\nprefix: /\nrewrite: /\nservice: uiLocal.kubeflow'
```

So the strong escaping appears to have changed and this appears to cause problems for Ambassador.

**What you expected to happen**:

I would expect it to always be

```
  annotations:
    getambassador.io/config: |-
      ---
      apiVersion: ambassador/v0
      kind:  Mapping
      name: centralui-mapping
      prefix: /
      rewrite: /
      service: centraldashboard.kubeflow
````

**How to reproduce it (as minimally and precisely as possible)**:

1. Create  a ksonnet app
1. Copy the files listed above to your components directory
1. ks show default -c test-central

**Anything else we need to know?**:

**Environment**:
- ksonnet version (use `ks version`):

```
ksonnet version: 0.11.0
jsonnet version: v0.10.0
client-go version: 
```
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.3", GitCommit:"2bba0127d85d5a46ab4b778548be28623b32d0b0", GitTreeState:"clean", BuildDate:"2018-05-21T09:17:39Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #682, opened 2018-07-02
Update ksonnet-lib printer to 0.1.10 to handle unescaping newlines

bryanl:
#670


Gopkg.toml
```diff
@@ -52,7 +52,8 @@ required = ["k8s.io/kubernetes/pkg/kubectl/cmd/util"]
 
 [[constraint]]
   name = "github.com/ksonnet/ksonnet-lib"
-  version = "v0.1.9"
+  source = "github.com/bryanl/ksonnet-lib"
```
shomron: We probably want to consume a new tag from the upstream repo.

PR #686, opened 2018-07-03
Create module when child exists

### LINKED ISSUE #675: Cant' create modules when a submodule exists
/kind feature

If you create a nested module with `ks module create a.b`, then you can't create module `a` at a later time.

bryanl:
Allow creation of module `a` when module `a.b` previously exists.
Changes module finding to look for `params.libsonnet` instead of
the module directory.

Fixes #675


bryanl:
@underrun I think it should error since the module doesn't exist. Could possibly cut down on errors.

Issue #688, opened 2018-07-03
Add --env flag to pkg install


shomron:
Package versions can be scoped to individual environments following the work in #631.
In this issue, we will add an `--env` flag to `ks pkg install` - which when specified, will scope the reference of the installed package/version to the requested environment.

Issue #626, opened 2018-06-19
Expose environment and version information in pkg list


shomron:
/kind feature

The `pkg list` command should expose package versions as well as environment mappings.

Possible output:

```
$ ks pkg list --installed
REGISTRY    NAME            INSTALLED    ENVIRONMENTS
========    ====            =========    ============
helm-stable spinnaker@0.2.4 *            production, dev
incubator   nginx@1.24.0    *
```

shomron:
Version column information added in #673

PR #697, opened 2018-07-05
Allow installing packages into environment scope

shomron:
* Add --env flag to pkg install cli
* Add support for environment qualifier in app.UpdateLib()
* Refactor CacheDependency to only focus on vendoring, app update is handled by the caller
* Move default app.Environment[s]() implementation to baseApp.


shomron:
Feature exercised in https://gist.github.com/shomron/828e3efa01039b80e3c23ca0b153eded

shomron:
@bryanl thanks for looking at this.

Regarding the initial notes:

* The behavior with late environment validation was documented [here](https://gist.github.com/shomron/828e3efa01039b80e3c23ca0b153eded#install-package-using-invalid-environment) - I don't think it's desirable, we can fix this here or as a followup PR.
* #626 documents the need to expose environment scope in `ks pkg list`. I'd like to handle this as a discrete task.

shomron:
Rebased before merge.

shomron:
Closes #688

Issue #615, opened 2018-06-15
ksonnet changes objects even in dry-run mode


bitfield:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
>
> /kind feature

**What happened**:

I ran `ks apply prod --dry-run`. Despite the `--dry-run` flag, ksonnet added labels and annotations to some cluster objects:

```
app.kubernetes.io/deploy-manager=ksonnet
ksonnet.io/managed={"pristine":"H4sIAAAAAAAA/3yPP0sEMRDFez/Gq+OtZyWpLS2sbOSK2eywBvNnyIwHuuS7SxQXRbgyvLzfb94GkvjETWMt8CARnc7HmY1u4fAaywKPe5ZU3zMXg0Nmo4
WM4DcUygyPudbMjcqK7qDCYWSNJcVACn90UE4crLYRZLLw8kAz...
```

**What you expected to happen**:

I expected no changes whatsoever to be made to the cluster in dry-run mode.

**How to reproduce it (as minimally and precisely as possible)**:

Do a `ks apply ENV --dry-run` to objects which have not been managed by ksonnet before.

**Anything else we need to know?**:

In:

```go
func (a *Apply) handleObject(co clientOpts, obj *unstructured.Unstructured) (string, error) {
	if err := tagManaged(obj); err != nil {
		return "", errors.Wrap(err, "tagging ksonnet managed object")
	}
```

it looks as if the call to `tagManaged()` should be wrapped with `if !a.DryRun { ... }`.

**Environment**:
- ksonnet version (use `ks version`):

```
ksonnet version: 0.11.0
jsonnet version: v0.10.0
client-go version:
```

- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.1", GitCommit:"d4ab47518836c750f9949b9e0d387f20fb92260b", GitTreeState:"clean", BuildDate:"2018-04-13T22:28:25Z", GoVersion:"go1.9.5", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #699, opened 2018-07-06
Don't change cluster in apply dry run modified

bryanl:
Add conditionals for object tagging and upserting.

#615


pkg/cluster/apply.go
```diff
@@ -176,7 +176,12 @@ func (a *Apply) handleObject(obj *unstructured.Unstructured) (string, error) {
 // preprocessObject preprocesses an object for it is applied to the cluster.
 func (a *Apply) preprocessObject(obj *unstructured.Unstructured) error {
 	dm := newDefaultManaged()
```
underrun: dryrun could exit earlier if we early out on a.DryRun before this line instead of doing this and then checking if we needed it.

PR #701, opened 2018-07-06
Don't install helm charts with hooks

### LINKED ISSUE #700: Don't install charts with helm hooks
/kind enhancement

Helm charts with hooks can't be handled with ksonnet's current flow since all components in a module are applied in order with no wait. In the first release of Helm chart support, reject any chart which contains hooks. Hooks are identified with an annotation `helm.sh/hook`.

bryanl:
If a helm hook is encountered in a file in a helm chart's `templates`
directory, return an error.

Fixes #700


bryanl:
@underrun There is no way to know unless you render the file. I checked the existing charts in stable and incubator, and none of them do that.

PR #708, opened 2018-07-09
Merge environment-scoped libraries with environment overrides

shomron:
This change allows environment-scoped libraries to be managed in the primary app configuration, while allowing overrides to be used for other environment attributes, such as server uri or k8s version.

### Test override pass-through of libraries

```
ks init pkg-upgrade && cd pkg-upgrade/

# Temporarily install in global scope to generate prototype (prototype commands are not yet env-scope aware)
ks pkg install incubator/nginx
ks prototype use io.ksonnet.pkg.nginx-simple myNginx

# Remove from global scope
sed -i -e  '/^libraries:$/,/version:/ d' app.yaml

# Install in stage scope only
ks env add stage
ks pkg install --env=stage incubator/nginx

# Add an override (w/o libraries)
ks env add --override stage --server https://localhost:8443

# Evaluation works for stage (libraries pass-through despite override)
ks show stage

# Show that server is indeed overridden
ks appy stage

# While we're here, show that default fails because the package is not scoped to that environment or global
ks show default
```


pkg/app/app001.go
```diff
@@ -54,17 +54,27 @@ func NewApp001(fs afero.Fs, root string) *App001 {
 
 // AddEnvironment adds an environment spec to the app spec. If the spec already exists,
 // it is overwritten.
-func (a *App001) AddEnvironment(name, k8sSpecFlag string, spec *EnvironmentConfig, isOverride bool) error {
+func (a *App001) AddEnvironment(e *EnvironmentConfig, k8sSpecFlag string, isOverride bool) error {
+	if a == nil {
```
bryanl: You don't have to check for this. It should panic if a is nil because it is a programming error.

shomron:
Removed receiver nil check in `AddEnvironment`

bryanl:
environments with overrides aren't annotated in `ks env list`

```
 $ cat app.override.yaml
apiVersion: 0.2.0
environments:
  default:
    destination:
      namespace: default
      server: https://localhost:6443
    k8sVersion: v1.10.3
    path: default
kind: ksonnet.io/app-override

 $ ks env list
NAME    OVERRIDE KUBERNETES-VERSION NAMESPACE SERVER
====    ======== ================== ========= ======
default          v1.10.3            default   https://localhost:6443
```

That override column should have a `*`.

bryanl:
This is an ambiguity that might need to be documented.

If you have an override for the `default` environment, and then you change the name of the environment with `ks set`, what happens to the override? The previous behavior would rename the override if there was one. 

```
$ ks env set default --name new-name
INFO Setting environment name from "default" to "new-name"
INFO Successfully moved "default" to "new-name"
$ ks env list
NAME     OVERRIDE KUBERNETES-VERSION NAMESPACE SERVER
====     ======== ================== ========= ======
default  *        v1.10.3            default   https://localhost:6443
new-name          v1.10.3            kafkatest https://localhost:6443
```

shomron:
> This is an ambiguity that might need to be documented.
> 
> If you have an override for the default environment, and then you change the name of the environment with ks set, what happens to the override? The previous behavior would rename the override if there was one.

This behavior I'm seeing on master is that the override is renamed and the original is not. I think we can do better, but I'll open a separate ticket.

shomron:
> environments with overrides aren't annotated in ks env list

Good call! I've pushed a fix for this.

shomron:
Squashed.

PR #711, opened 2018-07-09
Helm chart verification

bryanl:
Fixing regressions in the ksonnet helm integration:

* handle chart versions like `v0.1.3` when finding the latest version
* make kubernetes cluster version available to charts
* add ks helm repository verification tool

### DIFF
--- a/hacks/ks-verify-helm.rb
+++ b/hacks/ks-verify-helm.rb
@@ -1,0 +1,73 @@
+#!/usr/bin/env ruby
+
+# verify charts specified in a helm repository render with ks
+# usage: ks-verify-helm.rb [helm repo URL]
+
+require 'net/http'
+require 'open3'
+require 'timeout'
+require 'tmpdir'
+require 'uri'
+require 'yaml'
+
+DEFAULT_HELM_REPO_URL="https://kubernetes-charts.storage.googleapis.com/index.yaml"
+
+helm_repo_url = ARGV[0] || DEFAULT_HELM_REPO_URL
+
+tmp_dir = Dir.mktmpdir("ks-verify-helm-", "/tmp")
+at_exit { FileUtils.remove_entry(tmp_dir) }
+
+app_name = (0...6).map { (65 + rand(26)).chr }.join
+
+# generate a ksonnet app
+Dir.chdir(tmp_dir) do
+    `ks init #{app_name} --skip-default-registries > /dev/null 2>&1`
+end
+
+fails = []
+
+Dir.chdir("#{tmp_dir}/#{app_name}") do
+    uri = URI.parse(helm_repo_url)
+    response = Net::HTTP.get(uri)
+
+    `ks registry add helm #{helm_repo_url}`
+
+    repo = YAML.load(response)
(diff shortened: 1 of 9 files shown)


hacks/ks-verify-helm.rb
```diff
@@ -0,0 +1,73 @@
...
+require 'timeout'
+require 'tmpdir'
+require 'uri'
+require 'yaml'
+
+DEFAULT_HELM_REPO_URL="https://kubernetes-charts.storage.googleapis.com/index.yaml"
+
+helm_repo_url = ARGV[0] || DEFAULT_HELM_REPO_URL
+
+tmp_dir = Dir.mktmpdir("ks-verify-helm-", "/tmp")
```
shomron: You might use `tmp_dir = Dir.mktmpdir("ks-verify-helm-", Dir.tmpdir)`

`Dir.tmpdir` on my macOS does not resolve to `/tmp`.

hacks/ks-verify-helm.rb
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env ruby
```
shomron: Might be worthwhile to add a `.ruby-version` or a `Dockerfile` pointing to a supported ruby version.

pkg/helm/renderer.go
```diff
@@ -167,7 +179,19 @@ func (r *Renderer) renderWithHelm(componentName, raw, chartPath string) (map[str
...
-	vals, err := chartutil.ToRenderValuesCaps(c, config, options, nil)
+
+	kubeVersion, err := r.k8sVersion()
+	if err != nil {
+		return nil, errors.Wrap(err, "setting Kubernetes version for Helm")
+	}
+
+	caps := &chartutil.Capabilities{
+		KubeVersion: &version.Info{
+			GitVersion: kubeVersion,
```
shomron: Where does this ultimately get consumed? Are any of the other version-related fields in `version.Info` needed?

hacks/ks-verify-helm.rb
bryanl: It does, but I should use `Dir.tmpdir` if I ever wanted this to work in windows.

hacks/ks-verify-helm.rb
bryanl: It will work with any Ruby over 2.0, but the Dockefile idea is interesting.

hacks/ks-verify-helm.rb
shomron: Mine resolves to something like this:

```
/var/folders/gl/9th4vs815bv_bwgkxs9rmgr40000gn/T/ks-verify-helm-20180710-11142-19aege7
```

pkg/helm/renderer.go
bryanl: It is consumed in the helm chart templates. None of the other fields are consumed within any charts in the stable repo.

bryanl: Here's an example: https://github.com/kubernetes/charts/blob/74a23f7efbb332ea873df89f09d28654c48f6615/stable/cockroachdb/templates/_helpers.tpl#L5

PR #716, opened 2018-07-10
Support prototype resolution from environment-scoped packages

### LINKED ISSUE #713: Support prototypes from environment-scoped packages
Environment-scoped packages (see #697) containing prototypes should be compatible with the various `ks prototype` subcommands. Previously, only globally-scoped packages were supported.

shomron:
Closes #713

To exercise:

```
ks init test-713 && cd test-713/
ks env add stage
ks pkg install --env stage incubator/nginx
ks prototype list
ks prototype preview io.ksonnet.pkg.nginx-simple --name myNginx
ks prototype use io.ksonnet.pkg.nginx-simple myNginx
```


shomron:
Latest prototype resolution logic (selecting between multiple versions of an installed package) - only support semver at this time. A followup issue will be filed to add support for sorting GitHub revisions.

PR #725, opened 2018-07-11
Fix up for merging objects

bryanl:
* Process was using wrong modified document
* Added support for using open api schema
* Merger now can retrieve original object

**note** I'm not 100% sure how to test this, so I did a bunch of manual testing and it didn't seem to break anything.


pkg/cluster/tagger.go
```diff
@@ -16,13 +16,18 @@
...
 	"github.com/ksonnet/ksonnet/pkg/metadata"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 type annotationEncoder interface {
 	EncodePristine(map[string]interface{}) error
+	DecodePristine() (map[string]interface{}, error)
```
bryanl: Anyone have a better name for this interface?

pkg/cluster/merger.go
```diff
@@ -207,8 +216,11 @@ func (p *patcher) patchSimple(obj runtime.Object, modified []byte, source, names
 		return nil, nil, cmdutil.AddSourceToErr(fmt.Sprintf("serializing current configuration from:\n%v\nfor:", obj), source, err)
 	}
 
+	t := newDefaultManaged()
+	original, err := t.Fetch(p.mapping, obj)
+
 	// Retrieve the original configuration of the object from the annotation.
-	original, err := kubectl.GetOriginalConfiguration(p.mapping, obj)
+	// original, err := kubectl.GetOriginalConfiguration(p.mapping, obj)
```
shomron: Can the commented code be removed?

pkg/cluster/annotation.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (m *defaultAnnotationApplier) Fetch(mapping *meta.RESTMapping, obj runtime.Object) ([]byte, error) {
+	annotations, err := mapping.MetadataAccessor.Annotations(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	if annotations == nil {
+		return nil, nil
```
shomron: Can we document that a nil slice is a valid return value?

pkg/cluster/merger.go
bryanl: it should. good catch

PR #744, opened 2018-07-12
All user to force install package with --force

bryanl:
This feature is added to allow two versions of a package to be
installed. (e.g. having multiple versions of a helm chart installed)

cc: @GuessWhoSamFoo 

### DIFF
--- a/pkg/registry/cache.go
+++ b/pkg/registry/cache.go
@@ -30,7 +30,7 @@ import (
 // CacheDependency vendors registry dependencies.
 // TODO: create unit tests for this once mocks for this package are
 // worked out.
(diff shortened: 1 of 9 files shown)


pkg/registry/cache.go
```diff
@@ -93,7 +93,7 @@ func CacheDependency(a app.App, checker InstalledChecker, d pkg.Descriptor, cust
 	if err != nil {
 		return nil, errors.Wrapf(err, "checking package installed status")
 	}
-	if ok {
+	if ok && !force {
 		return nil, errors.Errorf("package '%s/%s@%s' already exists.",
 			libRef.Registry, libRef.Name, libRef.Version)
```
GuessWhoSamFoo: libRef.Version returns "" rather than SemVer; might be confusing later on

PR #618, opened 2018-06-16
WIP: add option to env set for k8s version

underrun:
enables --api-spec flag to `ks env set` command to enable changing the
kubernetes version for an environment

Open questions:
* i feel like i would prefer table based tests here - good time to move?
* these tests still feel a little clunky - suggestions on refactor if we move to tables?
* there are unused functions in pkg/actions (EnvSetNamespace EvnSetName) - old cruft? better direction?

PR #752, opened 2018-07-13
Add examples for CLI ref --api-spec and --server for env set

GuessWhoSamFoo:
XR #612 and #618

### DIFF
--- a/docs/cli-reference/ks_env_set.md
+++ b/docs/cli-reference/ks_env_set.md
@@ -25,9 +25,16 @@ ks env set <env-name> [flags]
 ### Examples
 
 ```
-#Update the name of the environment 'us-west/staging'.
+# Update the name of the environment 'us-west/staging'.
 # Updating the name will update the directory structure in 'environments/'.
 ks env set us-west/staging --name=us-east/staging
+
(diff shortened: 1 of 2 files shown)


docs/cli-reference/ks_env_set.md
```diff
@@ -25,9 +25,16 @@ ks env set <env-name> [flags]
...
-#Update the name of the environment 'us-west/staging'.
+# Update the name of the environment 'us-west/staging'.
 # Updating the name will update the directory structure in 'environments/'.
 ks env set us-west/staging --name=us-east/staging
+
+# Setting k8s API version for an environment
+ks env set us-west/staging --api-spec=version:v1.8.0
+
+# Updating the server
+ks env set us-west/staging --server=192.168.99.100:8443
```
underrun: this needs https:// i think (at least this would make it match what is by default generated when you crate an environment)

GuessWhoSamFoo: Yup, it requires `https://`. Probably could use validation or coercing here

GuessWhoSamFoo:
Addressed comment and squashed

PR #757, opened 2018-07-16
updating e2e tests/output

bryanl:
* package versions are SHA1 instead of tag name
* strip extra info off server versions before comparing
* ensure Kubernetes version is retained when updating environment namespace or name
* remove verbose travis output


bryanl:
This solution isn't going to work. I'm going to need to circle back aground on this version issue.

bryanl:
@ksonnet/reviewers I believe this works now and is ready to be reviewed.

PR #794, opened 2018-07-27
Allow ksonet type registry in an enterprise git

oazmon:
Out github is at github.intuit.com; Making this change allows ks to consider it a valid ksonet registry. We need this fix soon. In the long run I suggest a better detection method, such as uri connect to git or a helm service.

Signed-off-by: Omer Azmon <omer_azmon@intuit.com>


bryanl:
@oazmon you'll have to add a signed off by to your commit for us to accept this PR. 
https://github.com/ksonnet/ksonnet/blob/master/CONTRIBUTING.md

SEJeff:
Also, this should allow a custom domain. My `$EMPLOYER` uses github enterprise and we have a custom domain that doesn't include github. This should probably be configurable since the hostname of github enterprise is configurable per-company.

Issue #795, opened 2018-07-30
ks component rm doesn't work with module.component


bryanl:
> /kind bug

Given a module named `redis` inside module `tools`, you should be able to remove the component with `ks component rm tools.redis`. This does not work. (`ks component rm tools/redis` does work, but this behavior has been deprecated)

PR #796, opened 2018-07-30
Remove component in module with

bryanl:
Removes component in module with `ks component rm module.name`. The previous behavior did not respect the module name.

#795


shomron:
Looks good, no blockers.

There are a two edge cases we may want to address in followups -
  * Missing component file (`foo.jsonnet`) causes failure, but missing component reference in params.jsonnet does not. Do we want to gracefully recover so that when one or the other is out of sync the user can still remove the remaining parts?
  * If a component is ambiguous (`foo.jsonnet`, `foo.yaml` both exist), one will be arbitrarily removed and the other will remain

bryanl:
@shomron I created #799 to track the points you brought up.

PR #797, opened 2018-07-30
ks component list aggregates from all modules

### LINKED ISSUE #765: ks component list - module recursion support
`ks component list` only works with a single module at a time - this makes it difficult to get a full listing of application components and their structure within modules.

Rather than adding a `--recursive` flag, I propose making `ks component list` walk the module hierarchy and list all components by default.

shomron:
Fixes #765


bryanl:
@shomron There are no plans to remove `--module` from this is there?

pkg/pipeline/pipeline.go
```diff
@@ -128,7 +128,7 @@ func (p *Pipeline) Components(filter []string) ([]component.Component, error) {
 
 	components := make([]component.Component, 0)
 	for _, ns := range modules {
-		members, err := p.cm.Components(ns)
+		members, err := p.cm.Components(p.app, ns.Name())
```
bryanl: `module` used to be named `namespace`. That was changed for reasons, but there are still a bunch of `ns` littered through the code base. Would you mind cleaning up the ones in this PR?

shomron: Sure thing!

shomron:
> There are no plans to remove --module from this is there?

I wasn't planning on it - filtering still seems valuable. I'd like to think in future about consolidating all our filters under a shared `--filter` interface.

pkg/pipeline/pipeline.go
shomron: Pushed.

PR #805, opened 2018-07-31
Allow ksonet type registry in an enterprise git

oazmon:
Our github is at github.intuit.com; Making this change allows ks to consider it a valid ksonet registry. We need this fix soon. In the long run I suggest a better detection method, such as uri connect to git or a helm service.

Signed-off-by: Omer Azmon <omer_azmon@intuit.com>


oazmon:
@bryanl I updated with sig from other PR to this one

shomron:
@oazmon sorry for the delay and thank you for the contribution!

PR #817, opened 2018-08-06
Allow showing only env set parameters

### LINKED ISSUE #747: Listing parameters for environments should be able to list the values defined in the environment
/kind feature

Currently, `ks param list --env <env name>` lists all parameters known to an environment. It should be possible to list only the parameters set by the environment instead of the values inherited from components.

bryanl:
When listing parameters for an environment, add an option that only
shows parameters set in the environment itself. This will allow ks users
to show only the differences. The new option is `--without-modules`.

example:

```
 $ ks param list --env default
COMPONENT     PARAM   VALUE
=========     =====   =====
nested.redis3 envset  1
redis         envset  1
redis2        envset  1
redis2        name    'redis2'
redis2        values  {}
redis2        version '3.7.3'
$ ks param list --env default --without-modules
COMPONENT     PARAM  VALUE
=========     =====  =====
nested.redis3 envset 1
redis         envset 1
redis2        envset 1
```

Fixes #747


bryanl:
I'm going to take a deeper look at this. The module structure might have to be replicated internally rather than returning a `{}`.

PR #827, opened 2018-08-09
Remove override flag from ks env rm

### LINKED ISSUE #822: Operations should only affect overridden environment
If an existing env name is overriden with an identical name:

```
$ ks init app
$ ks env add default --override --server https://192.168.99.101:8443 --api-spec version:v1.8.0

# List env shows only one env
$ ks env list
NAME    OVERRIDE KUBERNETES-VERSION NAMESPACE SERVER
====    ======== ================== ========= ======
default *        v1.8.0             default   https://192.168.99.101:8443

# Deletes non-overridden default
$ ks env rm default
```

The user should not be able to delete the non-overridden environment.

`ks env rm default` should have deleted the overridden environment

**Environment**:
- ksonnet version (use `ks version`): 0.12.0
- Kubernetes version (use `kubectl version`):

GuessWhoSamFoo:
Closes #822 

This change ensures `ks rm` is consistent with `ks env set` with overriding.

 - `ks env rm` no longer has an `--override` flag
 - When overriding an env, it is not possible to change the non-overriden env.

### DIFF
--- a/e2e/env_test.go
+++ b/e2e/env_test.go
@@ -138,7 +138,7 @@ var _ = Describe("ks env", func() {
 					"default", e.serverVersion(), "default", "*", "http://example.com")
 				a.checkEnvs(expected)
 
-				o = a.runKs("env", "rm", "-o", "default")
+				o = a.runKs("env", "rm", "default")
 				assertExitStatus(o, 0)
 
 				a.checkEnvs(genEnvList(e.serverVersion()))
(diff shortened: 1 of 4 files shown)


GuessWhoSamFoo:
@bryanl `ks env rm` will default to the override if it exists.

GuessWhoSamFoo:
Closed per discussion on UX around this

PR #831, opened 2018-08-14
Add --tls-skip-verify global option affecting outgoing HTTP requests

shomron:
This change allows disabling TLS certification verification when interacting with GitHub/Helm registries. This is not recommended and is only provided as a option when working around trust issues with corporate proxies.

### Install and run mitmproxy:

```
brew install mitmproxy
mitmproxy
```

### Test registry / pkg commands via proxy, with and without verification

```
# registry add - fail
https_proxy=localhost:8080 ks registry add helm-stable https://kubernetes-charts.storage.googleapis.com
https_proxy=localhost:8080 ks registry add foo github.com/ksonnet/parts/tree/master/incubator

# registry add - success
https_proxy=localhost:8080 ks registry add helm-stable https://kubernetes-charts.storage.googleapis.com --tls-skip-verify
https_proxy=localhost:8080 ks registry add foo github.com/ksonnet/parts/tree/master/incubator --tls-skip-verify

# pkg list - fail
https_proxy=localhost:8080 ks pkg list

# pkg list - success
https_proxy=localhost:8080 ks pkg list --tls-skip-verify

# pkg install - fail
https_proxy=localhost:8080 ks pkg install incubator/nginx
https_proxy=localhost:8080 ks pkg install helm-stable/redis

# pkg install - success
https_proxy=localhost:8080 ks pkg install incubator/nginx --tls-skip-verify
https_proxy=localhost:8080 ks pkg install helm-stable/redis --tls-skip-verify
```


shomron:
I missed `ks init` - will push and update.

shomron:
@bryanl I'm going to merge this as-is, and continue the init work in a new issue. This change alone has value if one needed to work around network constraints when deploying an existing app.

bryanl:
@shomron This is fine, you always have the recourse of doing `ks init foo --skip-default-registries`

shomron:
Actually, there remains an issue with fetching the OpenAPI spec for ksonnnet-lib generation, so I think that we'll still need to account for that.

PR #833, opened 2018-08-15
Allow reuse of installed packages in other scopes

### LINKED ISSUE #728: Allow installing package when same version is installed in another scope
It should be possible to install a package into a scope, regardless of whether that version is also installed in another scope.

Currently this is a rejected with an error along the lines of:

```
ERROR package 'incubator/nginx@40285d8a14f1ac5787e405e1023cf0c07f6aa28c' already exists.
```

shomron:
Closes #728


shomron:
When installing a package that already exists into a scope (env or global) - it will now reference the previously vendored files and not require a `--force` flag. If `--force` is provided, the files will always be re-vendored.

PR #844, opened 2018-08-28
Fix sourcePath for imported json on root module

GuessWhoSamFoo:
This PR removes the extra `/` added because the default module is `/` since 0.11

Added test to hit the code path on https://github.com/ksonnet/ksonnet/blob/master/pkg/actions/import.go#L203

### DIFF
--- a/pkg/actions/import.go
+++ b/pkg/actions/import.go
@@ -271,9 +271,6 @@ func (i *Import) createComponentFromData(name, data string, templateType prototy
 
 func (i *Import) createComponent(fileName, base, ext string, templateType prototype.TemplateType) error {
 	var name bytes.Buffer
-	if i.module != "" {
-		name.WriteString(i.module + "/")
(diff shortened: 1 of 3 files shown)


pkg/actions/import.go
```diff
@@ -271,9 +271,6 @@ func (i *Import) createComponentFromData(name, data string, templateType prototy
 
 func (i *Import) createComponent(fileName, base, ext string, templateType prototype.TemplateType) error {
 	var name bytes.Buffer
-	if i.module != "" {
```
underrun: does this hold if someone specifies the "/" module explicitly (is that even allowed in this case?)

GuessWhoSamFoo: Yup, this should hold for cases with and without `--module /` passed from CLI. Default i.module is `/` if module is not specified explicitly.

Rebased with updated test cases

Issue #756, opened 2018-07-15
Run ksonnet without being a ksonnet application directory


bryanl:
/kind feature

It would be convenient if you could run actions on a ksonnet application without being in the ksonnet application directory. e.g.: `ks param set component key value --ks-root /app`. This would make it easier to update ksonnet parameters when working in another location (building an application)

PR #845, opened 2018-08-29
update cobra and pflag

underrun:
to use @shomron's fork via dep source magic

this will allow us to keep the imports as spf13 rather than using forks for a bunch of dependent packages.

also blocks #756 

Signed-off-by: Derek Wilson <derek@heptio.com>


shomron:
@underrun can you verify if you're using the latest dep release? It looks like there are differences in the lock file that may be related to that.

shomron:
:shipit:

Issue #836, opened 2018-08-20
ks CLI errors out when running --help (and other commands) outside a ksonnet application directory


underrun:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

running `ks --help` fails outside a ks application directory

**What you expected to happen**:

to see help

**Anything else we need to know?**:

This is a recent regression. Other commands that don't need the app directory also don't work.

**Environment**:
- ksonnet version (use `ks version`): as of git commit: 200abf0f
- Kubernetes version (use `kubectl version`): N/A

PR #847, opened 2018-09-02
fix -h, verison, and add global --dir flag with refactor

underrun:
this addresses both #756 and #836 but requires moving app loading to action handling rather than during cli command processing. 

there are still some messy bits and it could use tests before merge - but i'd like feedback on the refactor first. specifically is there a cleaner/better way to do this, and/or is there anything else along side app initialization that I could move to make command parsing and code execution more cleanly separated?


pkg/clicmd/init.go
```diff
@@ -174,22 +178,22 @@ func newInitCmd(fs afero.Fs, wd string) *cobra.Command {
 	return initCmd
 }
 
-func genKsRoot(appName, ksDir, wd string) (string, error) {
-	if ksDir == "" {
-		return "", errors.New("invalid working directory")
+func genKsRoot(appName, ksExecDir, newAppDir string) (string, error) {
```
shomron: A comment with examples of the different ways of calling would be helpful to me.

pkg/clicmd/prototype_use.go
```diff
@@ -84,34 +87,80 @@ ks prototype use single-port-deployment nginx-depl \
...
+		Long:    prototypeUseLong,
+		Example: prototypeUseExample,
+		FParseErrWhitelist: cobra.FParseErrWhitelist{
+			UnknownFlags: true,
+		},
 		DisableFlagParsing: true,
-		RunE: func(cmd *cobra.Command, rawArgs []string) error {
-			if len(rawArgs) == 1 && (rawArgs[0] == "--help" || rawArgs[0] == "-h") {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.DisableFlagParsing = false
```
shomron: Is there a way to avoid mutating the command here? Perhaps we can create a copy first?

pkg/clicmd/prototype_use.go
```diff
@@ -84,34 +87,80 @@ ks prototype use single-port-deployment nginx-depl \
...
+				if splitFlag[0] == flag.Name {
+					if len(splitFlag) == 2 {
+						// global flag with =
+						args = append(args[:i], args[i+1:]...)
+					} else if flag.NoOptDefVal != "" {
+						// optional flag that sets default
+						args = append(args[:i], args[i+1:]...)
+					} else {
+						// '--flag arg' means remove two
+						args = append(args[:i], args[i+2:]...)
```
shomron: Can this go out of bounds if args ends with `--flag` and no corresponding value?

pkg/clicmd/prototype_use.go
```diff
@@ -84,34 +87,80 @@ ks prototype use single-port-deployment nginx-depl \
...
 		},
 	}
 
 	return prototypeUseCmd
 }
+
+func removePersistentFlags(flags *pflag.FlagSet, args []string) []string {
+	flags.Visit(func(flag *pflag.Flag) {
+		for i, arg := range args {
+			if len(arg) == 0 || arg[0] != '-' || len(arg) == 1 {
```
shomron: `len(arg) < 2`?

pkg/clicmd/prototype_use.go
```diff
@@ -84,34 +87,80 @@ ks prototype use single-port-deployment nginx-depl \
...
+
+			if arg[1] == '-' {
+				if len(arg) == 2 {
+					// "--" terminates the flags
+					break
+				}
+
+				// long flag that might be known by ks
+				splitFlag := strings.SplitN(arg[2:], "=", 2)
+				if splitFlag[0] == flag.Name {
```
shomron: We may run into trouble if we ever use `FlagSet.SetNormalizeFunc`.

pkg/clicmd/prototype_use.go
```diff
@@ -84,34 +87,80 @@ ks prototype use single-port-deployment nginx-depl \
...
+			addGlobalOptions(m)
 
 			return runAction(actionPrototypeUse, m)
 		},
 	}
 
 	return prototypeUseCmd
 }
+
+func removePersistentFlags(flags *pflag.FlagSet, args []string) []string {
```
shomron: Can we utilize pflag to do the parsing? What about:
* One pass, parse with a flagset containing no flags, and `Unknown: true`. This should remove all flags and leave only positional arguments?
* Second pass, visit flags in original flagset. Construct new command line by appending all these with their values, plus the positional args.

Would that have the same outcome, without doing parsing logic on our end?

pkg/clicmd/prototype_use.go
underrun: heh - yeah - just borrowed some of this code from cobra itself because i wanted to make sure i'm parsing flags the same way they do. i'll tweak it.

pkg/clicmd/init.go
underrun: i will give an example or two, but i think the tests do a good job of showing all the cases?

pkg/clicmd/prototype_use.go
shomron: ^ reading that again, I think we would lose the unknown flags in the final output, which is not the intent.

pkg/clicmd/prototype_use.go
underrun: no - confirmed with `--tls-skip-verify` at end of command. but flow control wise we don't hit this else block unless we've been given a flag that requires an argument with its argument. because this runs after the original Parse we already know we have a valid command line apart from unknown flags (which we are ignoring anyway).

pkg/clicmd/prototype_use.go
underrun: actually it needs to be this way but i'll add a comment

pkg/clicmd/prototype_use.go
underrun: i'd really love not to copy pasta but i haven't found a good way out of it without modifying pflag

pkg/clicmd/prototype_use.go
shomron: What had bothered me was that a second execution would behave differently because of the mutation - largely academic given current application structure. Maybe just reset the flag the way you found it afterwards?

underrun: resetting it after sounds good actually - thanks for the suggestion

PR #855, opened 2018-09-07
Qualify libraries and refactor app

shomron:
Refactor schema to support explicit migrations

* Dropped support for 0.0.1 apps
* Versioning has been pushed up into the Schema types instead of App.
* Added migrations framework for migrating schema versions, one hop at time


pkg/app/app.go
```diff
@@ -95,42 +92,27 @@ type App interface {
...
 
 // Load loads the application configuration.
 func Load(fs afero.Fs, httpClient *http.Client, appRoot string) (App, error) {
-	log := log.WithField("action", "app.Load")
-
 	spec, err := read(fs, appRoot)
 	if os.IsNotExist(err) {
 		// During `ks init`, app.yaml will not yet exist - generate a new one.
-		return NewApp010(fs, appRoot, httpClient), nil
+		return NewBaseApp(fs, appRoot, httpClient), nil
```
underrun: could mock / override NewBaseApp to get some test coverage on `app.Load()` - not blocking just wanted to note it for future.

pkg/app/base_app.go
```diff
@@ -34,16 +37,47 @@ type baseApp struct {
...
+
+// optLoadFn overrides baseApp's load function, useful when testing.
+func optLoadFn(loadFn func() error) Opt {
+	return func(a *baseApp) {
+		a.load = loadFn
+	}
+}
+
+// optNopLoader overrides baseApp's loader to do nothing. (NOOP)
+func optNoopLoader() Opt {
```
underrun: why have this at all instead of just using the above `optLoadFn(func() error {return nil})` ... alternately  now that optLoadFn isn't called anywhere we could just delete that?

i guess really my assertion would be we only need one of `optLoadFn` or `optNoopLoader`.

pkg/app/migrations_test.go
```diff
@@ -0,0 +1,329 @@
...
+)
+
+func Test_migrateSchema010To020(t *testing.T) {
+	var src = &Spec010{
+		APIVersion:  "0.1.0",
+		Kind:        "ksonnet.io/app",
+		Name:        "test-migration",
+		Version:     "0.0.1",
+		Description: "test migration description",
+		Authors: []string{
```
underrun: can't approve of these test cases hard enough

shomron: 😊

PR #859, opened 2018-09-10
Add env targets cli docs; stringSlice -> string

GuessWhoSamFoo:
Ran make docs again

Added brief example usage docs for `ks env targets -h`

### DIFF
--- a/docs/cli-reference/ks_env_targets.md
+++ b/docs/cli-reference/ks_env_targets.md
@@ -1,27 +1,49 @@
 ## ks env targets
 
-Set module targets for an environment
+Set target modules for an environment
 
 ### Synopsis
 
-targets
+
+The `targets` command selects one or more modules to be applied by an
+environment. The default environment target is the root module, `/`.
+
+Changing targets for an environment will require specifying all desired modules including the root module.
+
 
 ```
 ks env targets [flags]
 ```
 
+### Examples
+
+```
+
+# Create a new module
+ks module create db
+
+# Generate a component and specify the module
+ks generate redis-stateless redis --module db
+
+# Change the default environment target from / to db
+# The targets are tracked in app.yaml
+ks env targets default --module db
+```
+
 ### Options
 
 ```
-  -h, --help                 help for targets
-      --module stringSlice   Component modules to include
+  -h, --help             help for targets
+      --module strings   Component modules to include
+  -o, --override         Set targets in environment as override
 ```
 
 ### Options inherited from parent commands
 
 ```
-      --tls-skip-verify      Skip verification of TLS server certificates
-  -v, --verbose count[=-1]   Increase verbosity. May be given multiple times.
+      --dir string        Ksonnet application root to use; Defaults to CWD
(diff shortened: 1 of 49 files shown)


pkg/clicmd/root.go
```diff
@@ -130,7 +130,7 @@ func NewRoot(appFs afero.Fs, wd string, args []string) (*cobra.Command, error) {
 	rootCmd.PersistentFlags().CountP(flagVerbose, "v", "Increase verbosity. May be given multiple times.")
 	rootCmd.PersistentFlags().Set("logtostderr", "true")
 	rootCmd.PersistentFlags().Bool(flagTLSSkipVerify, false, "Skip verification of TLS server certificates")
-	rootCmd.PersistentFlags().String(flagDir, wd, "Ksonnet application root to use; Defaults to CWD")
+	rootCmd.PersistentFlags().String(flagDir, "", "Ksonnet application root to use; Defaults to CWD")
```
underrun: i'm not sure you want this change in here?

GuessWhoSamFoo: So I kinda hacked around this - make the docs with an empty string to avoid having my local paths appear in the docs then rebuilt with the original default

docs/cli-reference/ks_env_targets.md
```diff
@@ -4,24 +4,43 @@ Set module targets for an environment
 
 ### Synopsis
 
-targets
+
+The `targets` command selects one or more modules to be select by an
```
shomron: typo `to be select`. Would it be clearer as _to be applied_ (by/in) an environment?

docs/cli-reference/ks_env_targets.md
```diff
@@ -4,24 +4,43 @@ Set module targets for an environment
...
+environment. The default environment target is the root path, `/`.
```
shomron: Root path -> root module

shomron: Should we mention that when changing targets, `/` will no longer be included implicitly?

shomron: Another potential callout - setting targets overwrites the entire list - they should always specify the full list of targets when updating.

pkg/clicmd/env.go
```diff
@@ -38,6 +38,7 @@ var (
 		"list":    "List all environments in a ksonnet application",
 		"rm":      "Delete an environment from a ksonnet application",
 		"set":     "Set environment-specific fields (name, namespace, server)",
+		"targets": "Set module targets for an environment",
```
shomron: target modules?

pkg/clicmd/env_targets.go
```diff
@@ -26,14 +26,32 @@ const (
 	vEnvTargetModules = "env-target-modules"
 )
 
+var (
+	envTargetLong = `
+The ` + "`targets`" + ` command selects one or more modules to be select by an
```
shomron: Typo as called out above.

shomron:
Looks good 👍. 

Minor comments inline. Please rebase and regenerate the docs to pick up recent changes.

PR #860, opened 2018-09-11
Handle override upgrades

shomron:

### DIFF
--- a/pkg/app/override.go
+++ b/pkg/app/override.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/blang/semver"
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


pkg/app/override.go
```diff
@@ -28,10 +29,13 @@ import (
 const (
 	// overrideKind is the override resource type.
 	overrideKind = "ksonnet.io/app-override"
-	// overrideVersion is the version of the override resource.
+	// OverrideVersion is the version of the override resource.
```
bryanl: The comment changed but the const did not. Was this intentional?

pkg/app/override.go
```diff
@@ -28,10 +29,13 @@ import (
...
 	overrideVersion = "0.3.0"
 )
 
+// OverrideSemVersion is a parsed semver.Version representing the current Override version.
+var overrideSemVersion = semver.MustParse(overrideVersion)
```
bryanl: The godoc doesn't match the variable.

pkg/app/override.go
shomron: No, there was a back and forth and I missed the comment when I changed it back. Thanks!

pkg/app/override.go
shomron: ...same issue as above.

PR #865, opened 2018-09-19
Add `--override` to env targets cli

### LINKED ISSUE #857: unify environment override behavior to be entirely --override flag based
currently some `ks env` commands will operate on `app.override.yaml` if it exists while other commands expect a `--override` flag to indicate whether we are operating on the base `app.yaml` env object or the `app.override.yaml` (if flag is present). 

we should move all env sub commands that modify either `app.yaml` or `app.override.yaml` to have the same behavior which should be to accept the `--override` flag.

underrun:
To bring them in line with env set and env add operations on override environments.

closes #857

### DIFF
--- a/pkg/app/base_app.go
+++ b/pkg/app/base_app.go
@@ -627,7 +627,7 @@ func (ba *baseApp) RenameEnvironment(from, to string, override bool) error {
 }
 
 // UpdateTargets updates the list of targets. Note this overrwrite any existing targets.
-func (ba *baseApp) UpdateTargets(envName string, targets []string) error {
+func (ba *baseApp) UpdateTargets(envName string, targets []string, isOverride bool) error {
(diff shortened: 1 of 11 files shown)


pkg/app/base_app.go
```diff
@@ -627,15 +627,15 @@ func (ba *baseApp) RenameEnvironment(from, to string, override bool) error {
...
+func (ba *baseApp) UpdateTargets(envName string, targets []string, isOverride bool) error {
 	spec, err := ba.Environment(envName)
 	if err != nil {
 		return err
 	}
 
 	spec.Targets = targets
 
-	isOverride := ba.IsEnvOverride(envName)
+	//isOverride := ba.IsEnvOverride(envName)
```
shomron: This can be removed.

PR #867, opened 2018-09-19
Sort resources semantically to tighten up reconciliation

### LINKED ISSUE #850: Sort Kubernetes resources according to known semantic dependencies
we can use the type of kubernetes resources we are working with to helps give us a better chance of successfully applying everything in one pass if we apply them in an order of mostly likely to have dependencies first (namespaces being the biggest example).

shomron:
Closes #850

### DIFF
--- a/pkg/cluster/sort.go
+++ b/pkg/cluster/sort.go
@@ -36,21 +36,27 @@ func (u UnstructuredSlice) Swap(i, j int) {
 }
 func (u UnstructuredSlice) Less(i, j int) bool {
 	// Ordered sort key extractors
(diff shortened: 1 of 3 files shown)


pkg/cluster/sort.go
```diff
@@ -64,15 +70,89 @@ func (u UnstructuredSlice) Less(i, j int) bool {
...
+	"StatefulSet",
+	"Job",
+	"CronJob",
+	"Ingress",
+	"APIService",
+}
+
+var kindOrderMap map[string]int
+
+func init() {
```
underrun: instead of having an init - could we use const + iota and just numeric sort?

PR #879, opened 2018-10-19
Support --kubeconfig flag for show command

aknuds1:
Support the `--kubeconfig` flag also for the `show` command.

I think the `ks show` command should also support the `--kubeconfig` flag, in addition to `$KUBECONFIG`, for completeness as I tend to be explicit and use `--kubeconfig` when I invoke `ks`. Other sub-commands to ks, such as `apply`, support `--kubeconfig` so I think it only makes sense that `show` support it too.


aknuds1:
@underrun Damn, I think you're right. Maybe I confused `ks show` with another subcommand. Still getting used to ksonnet :)

underrun:
No worries at all! Definitely appreciate your willingness to jump in with code; welcome to ksonnet :-)

